PR #100, opened 2014-11-14
Add ability to enter to values for compose object

yellowandy:
The method in the pull request is a little bit hackey and only works for the to field, but it does work.  Don't have time at the moment to pursue the other method of implementation which includes modifying the dom structure in several different places but wanted to pass this along in case it's useful at all.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1704,7 +1704,24 @@ var Gmail = function(localJQuery) {
       TODO: ability to set
      */
     to: function(to) {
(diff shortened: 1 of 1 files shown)


josteink:
@yellowandy Is this PR still relevant? Is it still current?

I'm closing stale issues, so if you think this should still go in, please re-open this issue and let me know and we'll take it from there.

Anyhow: Thanks for your contribution!

PR #140, opened 2015-02-12
Adding handling for show_newly_arrived_message

cmroberts:
It looks like the 'show_newly_arrived_message' event currently isn't being handled. This pull request addresses that issue. 

Detailed description of this fix: 
It turns out the response object was not being populated when the 'mo' event is being handled. This results in no event handlers being triggered, due to the conditional in line 745 that checks for a non-null response. It's entirely possible this event should be handled another way; however, I found the approach taken in this pull request to be reliable and useful. It reliably captures the id of the newly arrived message, along with the url and body. 

If you decide to go ahead with merging this pull request, the README should be updated to reflect that listeners for this event should only expect three parameters: the id, url, and body.

### DIFF
--- a/README.md
+++ b/README.md
@@ -756,12 +756,12 @@ gmail.observe.on("delete_forever", function(id, url, body, xhr) {
   console.log("id:", id, "url:", url, 'body', body, 'xhr', xhr);
 })
 
-gmail.observe.on("delete_message_in_thread", function(id, url, body, xhr) {
-  console.log("id:", id, "url:", url, 'body', body, 'xhr', xhr);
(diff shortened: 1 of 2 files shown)


KartikTalwar:
Hey @cmroberts nope, you're right. This is definitely a bug and your patch is most certainly the right way to fix it. As you already pointed out that the readme should be updated to reflect the new parameters. Would you mind adding that change and I'd be happy to merge :-)

cmroberts:
The README is updated! Thanks for confirming this approach and merging it back to master.

PR #145, opened 2015-02-27
get reply_to email

iblazevic:
Hi, I needed reply-to email for my extension so I got it this way.

I'm not sure its the best way tough cause [13][4] is an array and I'm not sure if reply to can be multiple addresses but I took into consideration only one. I couldn't find a way to get more as gmail and all other providers allow only one reply to.

Please take a look and let me know if there is a better way, I'll gladly update. This is something I see could be useful for lot of people.

Thanks!

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1555,6 +1555,19 @@ var Gmail = function(localJQuery) {
     }, 0);
   }
 
+  api.tools.get_reply_to = function(ms13) {
+    // reply to is an array if exists
+    var reply_to = (ms13 != undefined) ? ms13[4] : [];
+
+    // if reply to set get email from it and return it
+    if (reply_to.length !== 0) {
+      return api.tools.extract_email_address(reply_to[0]);
+    }
+
+    // otherwise return undefined
+    return;
+  }
+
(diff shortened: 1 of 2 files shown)


KartikTalwar:
Looking into this now

KartikTalwar:
hey @iblazevic this looks good. I commented on one small change in the code but if you can update the readme to show the new field and have that be part of the pull request, I'll merge it right away

src/gmail.js
```diff
@@ -1555,6 +1555,19 @@ var Gmail = function(localJQuery) {
...
+    // reply to is an array if exists
+    var reply_to = (ms13 != undefined) ? ms13[4] : [];
+
+    // if reply to set get email from it and return it
+    if (reply_to.length !== 0) {
+      return api.tools.extract_email_address(reply_to[0]);
+    }
+
+    // otherwise return undefined
+    return;
```
KartikTalwar: can this explicitly return `null`

iblazevic:
hi @KartikTalwar I updated the readme and your comment, let me know if something else needs to be changed

Thanks!

PR #147, opened 2015-03-02
Update README.md

Projjol:
Reckoned an example was needed there.


KartikTalwar:
@Projjol the new CSP prevents injecting from the URL hence the comment to say 'copy paste gmail.js source code'. That codeblock in the readme is intentional

PR #151, opened 2015-03-03
Check that email is visible when looking up email_id in preview pane mode

danielbreves:
`gmail.get.email_id()` returns the id for the wrong email when in preview_pane mode. This seems to happen because [`api.dom.email_contents()`](https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js#L267) returns all emails that have been viewed by the user when in preview pane mode and `email_id` only [returns the first one](https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js#L280).


atroche:
awesome!

princemaple:
@KartikTalwar

KartikTalwar:
Thanks for catching that @danielbreves

PR #165, opened 2015-04-17
Added async versions for make_request, visible_emails, email_data and email_source

mzohreva:
I have made asynchronous versions of the following functions:
* `make_request`
* `visible_emails`
* `email_data`
* `email_source`

The async versions have the postfix **_async** in their names and take an additional argument called **callback** which will be called when the results are ready. The callback is passed a single argument which is the data that is normally returned by the synchronous version of the function.

I checked the code and found that the following functions depend on `make_request` either directly or indirectly:
* `api.get.visible_emails`
* `api.get.email_data`
* `api.get.email_source`
* `api.get.selected_emails_data`
* `api.get.displayed_email_data`
* `api.dom.email.prototype.data`
* `api.get.email_ids`
* `api.dom.email.prototype.source`
* `api.check.is_thread`

I have added async versions for the first 3 functions listed above and if people like the idea, they can add async versions for the rest as well. I really only needed `email_data` for my own project.

Using the async versions improves the responsiveness of the user interface, especially if you use any of these functions in event handlers for events like *view_thread* or *view_email*

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -22,6 +22,17 @@ var Gmail = function(localJQuery) {
               compose : {}
             };
 
+  // The helper object is private
+  var helper = {
+                  get : {},
+                  observe : {},
+                  check : {},
+                  tools : {},
+                  tracker : {},
+                  dom : {},
+                  chat : {},
+                  compose : {}
+                };
 
   api.version           = "0.4";
   api.tracker.globals   = GLOBALS;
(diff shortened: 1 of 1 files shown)


KartikTalwar:
Hi @mzohreva, thanks for taking the time to handle the async case!

Although it might be time to allow aync xhr requests I don't think duplicating the code is the best way to go about it. Instead of having 2 functions with the exact same logic, it might be best to separate out the "processing" part of the function as a helper. Do you think we might be able to add the async functions without writing the logic twice?

mzohreva:
Good point! I added a new commit to solve the problem.

mzohreva:
@KartikTalwar, did you see my new changes?

onassar:
Would love to make use of this pull if it's up-to-par

KartikTalwar:
I'm reviewing a few outstanding issues so hopefully this will be merged
tonight. Sorry for the delay

onassar:
Np @KartikTalwar :)

PR #175, opened 2015-05-12
Accommodate pagination

onassar:
eg. https://mail.google.com/mail/u/0/#trash/p2
Should work properly now


onassar:
Ahh, this problem is a bit more complicated. Apologies

PR #176, opened 2015-05-13
Method updates + new page number method

onassar:
- Changed the way `current_page` works to include many edge-cases (eg. compose window being open, results being paginated, etc)
- Added `current_page_number` to retrieve page-number details
- Added `emails_per_page` to retrieve the number of emails shown per-result
- Added an `http_event` observer, to update the tracker details when the number of emails per page is changed by the client in their settings area
- Updated `visible_emails_pre` to properly request only the number of emails required, and to start the request at the correct offset, based on the page-number the client is currently on (eg. `/p2`, `/p3`, etc.)

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1,3 +1,4 @@
+
 var Gmail = function(localJQuery) {
 
   /*
@@ -1418,12 +1419,16 @@ var Gmail = function(localJQuery) {
 
 
   api.helper.get.visible_emails_pre = function() {
-    var page = api.get.current_page();
-    var url = window.location.origin + window.location.pathname + '?ui=2&ik=' + api.tracker.ik+'&rid=' + api.tracker.rid + '&view=tl&start=0&num=120&rt=1';
-    
+    var page = api.get.current_page(),
+      limit = api.get.emails_per_page(),
+      page_number = api.get.current_page_number(),
+      start = page_number * limit;
+    var url = window.location.origin + window.location.pathname + '?ui=2&ik=' + api.tracker.ik+'&rid=' + api.tracker.rid + '&view=tl&start=' + (start) + '&num=' + (limit) + '&rt=1';
+
     if(page.indexOf('label/') == 0) {
       url += '&cat=' + page.split('/')[1] +'&search=cat';
     } else if(page.indexOf('category/') == 0) {
+      var cat_label;
       if(page.indexOf('forums') != -1) {
         cat_label = 'group';
       } else if(page.indexOf('updates') != -1) {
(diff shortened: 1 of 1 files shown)


KartikTalwar:
can you add documentation for the new functions to the readme please

onassar:
Sure. Didn't want to go ahead with it unless it looked good for inclusion.
Want to merge this and then I'll do a separate pull for the readme?

src/gmail.js
```diff
@@ -1505,23 +1510,68 @@ var Gmail = function(localJQuery) {
...
+   */
   api.get.current_page = function() {
-    var hash  = window.location.hash.split('#').pop().split('?').shift();
-    var pages = ['sent', 'inbox', 'starred', 'drafts', 'imp', 'chats', 'all', 'spam', 'trash', 'settings'];
-
-    var page = null;
-
-    if($.inArray(hash, pages) > -1) {
-      page = hash;
+    var pages = ['sent', 'inbox', 'starred', 'drafts', 'imp', 'chats', 'all', 'spam', 'trash'];
```
KartikTalwar: why id `settings` not part of the list anymore?

onassar: Since `settings` & `settings/*` can't contain any threads (eg. `settings/1748vsbdnd`), it's returned through the rest of the conditions/logic in the function. This function is only used to filter the regular expression to pages that could possible have a thread as part of the hash.

src/gmail.js
```diff
@@ -2390,5 +2440,19 @@ var Gmail = function(localJQuery) {
     return false;
   }
 
+  /**
+   * Track the change of emails per page
+   * 
+   */
+  api.observe.on(
```
KartikTalwar: I think it would be best to not activate any observer methods unless the user starts it. Hijacking http calls should be necessary on when its needed by the user. You should clarify that this needs to run along with the get emails endpoint.

KartikTalwar:
@onassar it'd be convenient to have the documentation changes in the same pull request

src/gmail.js
KartikTalwar: thanks for clarifying

src/gmail.js
onassar: Totally get ya. But what's the best way forward, here? `emails_per_page` is used by `visible_emails_pre`. If the settings are changed, this detail needs to be captured from `GLOBALS`

I'm actually doing something a bit more radical that's not in the pull-request:
Because Gmail cache's listings on the client-side, I actually reload Gmail when the settings are changed and it's for the number of emails per page. I naturally didn't include that in the PR, but I couldn't find a way to release the email-listing-caches (eg. starred-view, label-view, etc) to ensure the counts are right.

onassar:
Gotcha

josteink:
I'm trying to clean out dead issues, and right now this one seems to have gone 100% stale. I'm therefore closing it until I hear otherwise.

@onassar Are you still interested in landing this PR? Are these patches still current or relevant?

Please let me know, by re-opening the issue and chiming in and then we can take it from there.

Issue #194, opened 2015-08-27
Bug - Gmail.js not defined in "new message" only window


JerryBels:
Hello !

So I found a bug that's really annoying. If someone ctrl+clic on the "new message" button of gmail, it opens a new tab containing the form to send the mail.

But, in that page, Gmail.js doesn't work. In my own plugin I'm working on, it says "GLOBALS is not defined", and even the boilerplate doesn't work at all, showing "Gmail is not defined".

To reproduce, simply : 
1/ Grab a fresh install of https://github.com/KartikTalwar/gmail-chrome-extension-boilerplate
2/ Add to it latest https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js
3/ Add it to Chrome
4/ Open your Gmail tab. The console shows "Hello xxx@gmail.com" as expected.
5/ Ctrl+clic on the "new message" big red button
6/ On the newly opened tab, open console to see the error message, and no hello :(

KartikTalwar:
Hey @JerryBels, this is expected since `GLOBALS` is the js variable that stores a few key user info in gmail.com which this library accesses. A popup from either a compose/print/etc does not contain the variable therefore the initialization fails. Would it be possible for you to describe what your extension wants to do in the popup window? There are a few things that you'd definitely not be able to do but I can look into adding an override

JerryBels:
Hey @KartikTalwar, thanks for answering !

Well, we're doing many things in our extension, using Gmail.js, that we need when a new compose window is opened. Detecting the user's email is the first. Then : 

    gmail.observe.on("compose")

is used to inject a toolbar.

gmail.observe.before('send_message')

is used to add elements to the body of the message.

provider.observe.after('send_message')

is used to get the ID of the mail sent and send it over Ajax to our server.

Currently that's all we do, but I guess that would be great to have this tab working like every other, although I realise now this could be a bit tricky since Gmail doesn't work the same in that tab, since it's more like a popup...

Thanks again !
Jeremy

KartikTalwar:
@JerryBels upon closer inspection, if you send a message from the popup window, the action still gets transferred back to the main gmail window which makes the request. This means that the observer will fire in the main window and should not require re-initialization in the popup window.

Because those `compose` and `send_message` observers don't require any `GLOBALS` properties, can you try changing line 27-31 in your gmail.js to these. Hopefully this will make everything work

```js
  api.version           = "0.4";
[...]

PR #199, opened 2015-10-26
Handle using the popup "new message"

JerryBels:
When clicked with ctrl button pushed, the "new message" button opens a popup. This popup doesn't work with gmail.js functionnalities.

This fix make it work with it. See https://github.com/KartikTalwar/gmail.js/issues/194


KartikTalwar:
@JerryBels thank you so much for fully investigating and figuring out the solution!!

PR #207, opened 2015-11-12
add a MutationObserver to handle and trigger recipient_change event o…

### LINKED ISSUE #119: observe.on recipient_change does not trigger when deleting recipient
The recipient_change triggered when adding recipient, but not trigger when deleting.

yellow-beard:
…n receipient remove from compose interface - fix #119

This gets the basic functionality that I need working, but guessing you might have suggestions for changes?

Don't really like listening on such a high level element, but not sure what would be better? Should I just register on the `document` instead?

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1285,11 +1285,10 @@ var Gmail = function(localJQuery) {
             var removedNodes = mutation.removedNodes;
             for (var j = 0; j < removedNodes.length; j++) {
               var removedNode = removedNodes[j];
-              console.log(removedNode);
               if (removedNode.className == 'vR') {
                 var observer = api.tracker.dom_observer_map['vR'];
                 var handler = api.tracker.dom_observers.recipient_change.handler;
-                api.observe.trigger_dom(observer, $(removedNode), handler);
+                api.observe.trigger_dom(observer, $(mutation.target), handler);
               }
             }
           }
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1277,6 +1277,25 @@ var Gmail = function(localJQuery) {
         $(window.document).bind('DOMNodeInserted', function(e) {
           api.tools.insertion_observer(e.target, api.tracker.dom_observers, api.tracker.dom_observer_map);
         });
+
+        // recipient_change also needs to listen to removals
+        var mutationObserver = new MutationObserver(function(mutations) {
+          for (var i = 0; i < mutations.length; i++) {
+            var mutation = mutations[i];
+            var removedNodes = mutation.removedNodes;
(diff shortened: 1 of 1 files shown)


src/gmail.js
```diff
@@ -1277,6 +1277,24 @@ var Gmail = function(localJQuery) {
...
+              var removedNode = removedNodes[j];
+              if (removedNode.className == 'vR') {
+                var observer = api.tracker.dom_observer_map['vR'];
+                var handler = api.tracker.dom_observers.recipient_change.handler;
+                api.observe.trigger_dom(observer, $(mutation.target), handler);
+              }
+            }
+          }
+        });
+        mutationObserver.observe($('div.dw')[0], {subtree: true, childList: true});
```
KartikTalwar: @yellow-beard `$('div.dw')` does indeed seem a very high level element to observe. Since the change is only to be found in the compose window, the mutation observation can be narrowed down to `$('.AD')`

yellow-beard: The issue is that I'm creating the observer when the script loads and before compose windows are opened, so it needs to be on something high level. To add it to compose windows, I'd need to add a new observer to each compose window when it is opened, which seems a little weird / cumbersome. I've also noticed the script doesn't always catch compose open events, if I click the compose button very quickly after gmail loads, or if I refresh a page with a compose window already open.

yellow-beard:
I've actually moved the observer to listen on the document.body, because I was seeing occasional errors listening on the element in the DOM which wasn't always rendered when gmail.js was loading.

KartikTalwar:
@yellow-beard that's alright - it can observe the entire body, in that case, could you ensure that on `api.observe.off()` the mutation observer gets disabled appropriately. You can use the `api.tracker` dict to keep track of the observer reference.

yellow-beard:
I'm a little confused sorry.

The mutation observer will be created once, just after the `window` `DOMNodeInserted` listener is created.

If at some later time I remove the mutation observer, new listeners binding with `gmail.observe.on('recipient_change'...` will not be triggered on recipient removes.

I've checked that if you bind an observer with `gmail.observe.on('recipient_change` and then call `gmail.observe.off('recipient_change')`, the event is no longer triggered when removing a contact. Perhaps this is what you mean in your comment? There is no work needed to make this happen, looks like the existing code takes care of it.

Tangent: I can't seem to bind multiple callbacks to 'recipient_change'. Is this intended / known?

KartikTalwar:
Weird, after running `.off()` the callback still fired for me, I'll double check and see if it happens again

KartikTalwar:
All good!

yellow-beard:
Awesome, thanks for the merge! Totally stoked to contribute!

PR #183, opened 2015-06-12
Fixed issue where HTML Comments were causing error spam in api.tools.insertion_observer

xueye:
HTML Comments are weird in that they have the attribute for `className`, but it's undefined. This PR fixes an issue where HTML comments being inserted into DOM would cause errors to appear in the console.

PR #219, opened 2015-12-18
Fix `Uncaught TypeError` in api.tools.insertion_observer

### LINKED ISSUE #218: gmail.js throws errors when installed with SalesforceIQ Chrome Extension
When both a gmail.js Chrome Extension that listens for DOM events and the SalesforceIQ Chrome Extension are installed, gmail.js's `api.tools.insertion_observer` throws many errors.

Steps to reproduce:

1. Install the [Salesforce IQ Chrome Extension](https://chrome.google.com/webstore/detail/salesforceiq/cnkgdfnjmgamkcpjdljdncfjcegpgcdg?hl=en-US)
1. Clone the [gmail.js Chrome Extension Boilerplate](https://github.com/KartikTalwar/gmail-chrome-extension-boilerplate)
1. Update gmail.js boilerplate the with the latest gmail.js (commit 3907fce538d6be1b50461893ee5b0befafe51718)
1. Change the main function in main.js to listen for `on compose` events:

    ```javascript
    var main = function(){
      // NOTE: Always use the latest version of gmail.js from
      // https://github.com/KartikTalwar/gmail.js
      gmail = new Gmail();
      console.log('Hello,', gmail.get.user_email());
      gmail.observe.on('compose', function(compose) {
        console.log('open compose', compose);
      });
    }
    ```

1. Install the gmail.js boilerplate extension in Chrome
1. Refresh Gmail

In the console, this will result in many errors of the form:
```
Uncaught TypeError: Cannot read property 'trim' of undefined              gmail.js:1335
```

Putting in a conditional breakpoint on that line reveals that the nodes causing this problem are `comment` nodes with a value of `ngIf: !!isShowing()`.  These nodes do not have a className property and so Gmail.js is breaking.

The fix is simple.  I'll be submitting a pull request with it shortly.

sbrudz:
When the target node has no className property, such as for a `comment` node, gmail.js will throw an `Uncaught TypeError`.  This commit guards against that case.

[Fixes #218]

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1332,7 +1332,8 @@ var Gmail = function(localJQuery) {
     if(!api.tracker.dom_observer_map) return;
 
     // loop through each of the inserted elements classes & check for a defined observer on that class
-    var classes = target.className.trim().split(/\s+/);
+    var className = target.className;
+    var classes = className ? className.trim().split(/\s+/) : [];
     if(!classes.length) classes.push(''); // if no class, then check for anything observing nodes with no class
     $.each(classes, function(idx, className) {
       var observer = dom_observer_map[className];


sbrudz:
I just noticed PR #183 which also addresses this problem.

KartikTalwar:
@sbrudz thanks!

PR #222, opened 2015-12-31
Change is_inside_email to reflect the changes on 8d74d82

NicoSantangelo:
After 8d74d82 `current_page` returns `'email'` instead of `null` if the user is inside an email


KartikTalwar:
@NicoSantangelo thanks for catching this!

PR #225, opened 2016-01-05
GLOBALS and VIEW_DATA undefined error fix

itsazzad:


KartikTalwar:
@itsazzad before I merge this, can you tell me when you are seeing the undefined error? When does it happen, what pages does it happen on? I'd like to reproduce it

itsazzad:
It happened on GMail page load.
screen shot 2016-01-05 at 11 21 03 pm

KartikTalwar:
Would it be possible to see how the chrome extension is being loaded? To me this sounds like `Gmail()` functions are being called before the page is completely loaded..

itsazzad:
I started with https://github.com/KartikTalwar/gmail-chrome-extension-boilerplate

itsazzad:
So https://github.com/KartikTalwar/gmail-chrome-extension-boilerplate/blob/master/main.js is making it

itsazzad:
I have found the issue.
If the extension is not restricted to GMail only then the errors coming.
So my suggested fork is suitable if in need of message passing with another domain.

KartikTalwar:
Thanks for giving those details. Since it is not recommended to run/include the js file for other domains in your manifest, for now I will close this pull request.

PR #244, opened 2016-03-14
Credit where credit is due!

yoiang:
I've run into gmail.js used in other projects with varying methods of credit. This makes it difficult keep up to date as well!

ps: great library @KartikTalwar :)


KartikTalwar:
Thanks @yoiang!

PR #252, opened 2016-05-03
Fixing undefined error if no data retrived

jmchia:
Fixed api.helper.get.email_data_post


src/gmail.js
```diff
@@ -1776,6 +1776,10 @@ var Gmail = function(localJQuery) {
 
 
   api.helper.get.email_data_post = function(get_data) {
+    
+    if (get_data.length == 0)
```
KartikTalwar: could you add braces for the if statement please

PR #253, opened 2016-05-04
Fixing undefined error if no data retrived

jmchia:
Fixing GET request parser when no data retrieved

will evaluate to true if value is not:

- null
- undefined
- NaN
- empty string ("")
- 0
- false


KartikTalwar:
thanks!

PR #256, opened 2016-05-13
Gmail updated DOM location of email id

### LINKED ISSUE #255: Cannot read property 'pop' of null
I got this error:
gmail.js:2342 Uncaught TypeError: Cannot read property 'pop' of null

I write this quick fix:
```js
if(this.id_element.attr('class').match(/(^|\s)m([\S]*)/) != null)
        this.id = this.id_element.attr('class').match(/(^|\s)m([\S]*)/).pop();
```

brentkelly:
Fix to find id of an email in the DOM tree after gmail updates.

Fixes #255

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1814,7 +1814,7 @@ var Gmail =  function() {
   api.dom.email = function(element) {
     if (typeof element == 'string') {
       this.id = element;
-      this.id_element = $('div.ii.gt.m' + this.id);
+      this.id_element = $('div.ii.gt div.a3s.aXjCH.m' + this.id);
       element = this.id_element.closest('div.adn');
     } else {
       element = $(element);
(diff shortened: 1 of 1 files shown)


nathan-muir:
I noticed a few other breakages when patching this one.

Fixes can be found here: https://github.com/nathan-muir/gmail.js/commit/405682538b0d2a6bd5db297578e1e2abf2b0f7be

YesThatAllen:
I applied these to lines 2332 and 2341 in master today, and it fixes https://github.com/KartikTalwar/gmail.js/issues/275

steelx:
Im still not able to catch any events on Observe method.

josteink:
Can you confirm if this works without these changes?

Maybe we need to look in both locations?

PR #261, opened 2016-05-20
email_id returns an empty string in a message popup window

koblucki:
When a message is popped out using the ‘In new window’ icon located next to the ‘Print all’ icon adjacent to the subject line the email_id isn’t calculated properly


KartikTalwar:
Sorry for the delay!

PR #262, opened 2016-05-26
Fix not getting the correct id element.

ronoudgenoeg:
Gmail seems to have changed where the id element is, giving a cannot read property pop of element null error with the old selector.


KartikTalwar:
Sorry for not responding to this sooner but I believe another pr has been merged that addresses this issue. Please let me know if I'm mistaken

PR #266, opened 2016-06-07
Fix query failure if chat is disabled

shanemcgraw:
When querying for the chat box in order to report ```is_right_side_chat```, a ```.getAttribute``` lookup on the query will evaluate to undefined, causing an error.

The error can be circumvented by checking the length of the chat box query first, and if it's disabled,  return false immediately.


KartikTalwar:
Sorry for the delay

Issue #267, opened 2016-06-08
gmail.get.selected_emails_data() is not working for list of emails from custom folders and search page


mayurizingadeCL:
Hello,

I am getting bellow error from gmail.js, when I am in inbox and added some search criteria.(search page).

SyntaxError: expected expression, got '<'.

Findings : 
I found, this method internally calls "visible_emails_pre" method.
In that I am getting url as "https://mail.google.com/mail/u/0/?ui=2&ik=6fdb4f9bc7&rid=mail%3Ase.3468.1.1&view=tl&start=0&num=120&rt=1&search=search". Its not working on browser tab as well, throwing 404 error code.

Please provide me the solution.

Thanks,
Mayuri Zingade

mayurizingadeCL:
Thanks,
It worked.

PR #268, opened 2016-06-09
Include at parameter in visible_emals_pre

seepel:
It appears that there is a parameter (`at`) missing when requesting email data from search. Not too familiar with this project, this seems to work for me but it very well could break other things or may not be the correct solution. I believe this is the same as #267


KartikTalwar:
Sorry for the delay in merging this

Issue #270, opened 2016-06-12
TypeError: Cannot read property 'GLOBALS' of null


foligny:
I got a GLOBAL error ! ;)
Well it`s the first time I try it, I might have setup things not correctly.
I get this error. Any idea if it's my fault or not?
On Chrome version 51.0.2704.84 m
![gmail js-error-global](https://cloud.githubusercontent.com/assets/7926543/15988809/2c6e5274-302c-11e6-988b-6283af1da99a.png)

foligny:
BTW you know you don't need to test both 
`typeof window.opener != 'undefined' && window.opener != null`
since you use != and not !==
Just 
`window.opener != null`
Tests for both undefined and null
Check out my blog on the most common JS errors : http://tech.yipp.ca/node-js/most-common-javascripts-errors-not-to-do/

foligny:
Actually I'm probably not using it correctly because 
`var email = gmail.get.user_email();`
returns undefined
Any idea why it could fail ?

JerryBels:
Hey, have you used https://github.com/KartikTalwar/gmail-chrome-extension-boilerplate ? If not I strongly advise you to take a look at it.

foligny:
Hi, thanks for the quick followup !
Yes I had seen it and had originally built my manifest.json based on it.
But trying it out by itself did help me find out what was wrong by getting one project with the boilerplate and add-in my code little by little until it breaks. It turns out it was because I was using **scriptTagContext.js** which injects all new script in a different execution environment (https://github.com/Rob--W/chrome-api/tree/master/scriptTagContext) so the script doesn't have access to gmail global vars.. duh!
Now it works by removing scriptTagContext.js
BTW my pull request does fix the if (.. != "undefined") test which doesn't do the job and throws an exception when the variables are not found since window.opener does not exists when it tries to access window.opener.GLOBALS

Thanks,
Francois

PR #271, opened 2016-06-12
Fix issue https://github.com/KartikTalwar/gmail.js/issues/270

foligny:


foligny:
It also includes the fix for the issue #270 and #273

JerryBels:
Nice little fixes, thanks.

Issue #294, opened 2016-08-16
gmail.get.email_data() plain_content/content_html fields return truncated email body


davidsertillange-optimizely:
Hi there, 

When getting the email content using gmail.get.email_data(email_id), the body of the emails (obtained through the plain_content or content_html fields) are truncated. Three dots (...) are added after the first 75 characters or so. 

Is this expected behavior? 

Best,
David

andy-sal:
We have been experiencing this issue as well. It appears the truncation is correlated with the is_deleted flag (true results in truncation).

davidsertillange-optimizely:
Check pull request #296 for a fix to this issue.

andy-sal:
Unfortunately it doesn't seem to have fixed my issue. I have tried calling get.email_data_full both with and without specifying an email id and it returns an identical object to get.email_data (with the same emails truncated to 77 characters followed by three dots(...)). As mentioned earlier it seems to have some association with the is_deleted flag. I have also found that if you star the email and call the method again it returns the full content.

michi88:
this is working for me: https://github.com/KartikTalwar/gmail.js/pull/329

josteink:
With [this PR](https://github.com/KartikTalwar/gmail.js/pull/329) now merged, could you check if this is still an issue?

If not, this issue probably should be closed.

josteink:
@davidsertillange-optimizely: Care to check if your issue is solved now?

josteink:
Since I haven't heard anything new on this issue, I'm closing it.

Please re-open it if you still have an issue you need resolved.

ShivanKaul:
I seem to be getting a truncated content_html as well, similar to what @davidsertillange-optimizely initially reported. The email is not deleted.

![image](https://cloud.githubusercontent.com/assets/5284154/25689018/f0bfcb96-3051-11e7-9558-0c00984ebbd9.png)

josteink:
Can you provide any more details on how to reproduce this? Things which could help:

* Sample code
* Gmail account-type (@gmail.com, Google Apps/GSuite-account, Grandfathered account from Google apps beta, etc)
* Use-case scenario (In a email-thread, or outside?)

Unless we can find a way to reproduce it, chances of getting it fixed is effectively zero.

ShivanKaul:
Yes, of course.

**Steps To Reproduce:**

1. Open `mail.google.com` (account is @gmail.com)
[...]

PR #296, opened 2016-08-18
Fetch full message body when getting email data

4i60r:


davidsertillange-optimizely:
That was exactly what I was looking for and fixes issue #294. Thanks a lot!

README.md
```diff
@@ -373,10 +376,20 @@ the data for the specified id is returned instead of the email currently visible
...
+
+Same as above, but returns full HTML email body for each message in thread.
+(Sends two requests instead of one)
+
+
 #### gmail.get.email_data(email_id=undefined, callback)
 
+Does the same as above but accepts a callback function
+
+#### gmail.get.email_data_full(email_id=undefined, callback)
```
josteink: Shouldn't this be get.email_data_full_async ?

josteink:
Sorry about the extremely slow response.

Apart from the comment about the documentation, it seems from the linked issue that this commit may not reliably fix this issue.

As such, having some clear instructions on how to reproduce the error-condition, which this PR could be tested against would be nice.

(I'm saying this, because for me email_data() and email_source() gets me all the stuff I need)

andy-sal:
Unfortunately I don't have any clear instructions but for me, it seemed to be associated with the is_deleted flag. It appeared to truncate the messages to 80 characters when it was set to true (even though emails weren't deleted). From memory it was only in large email threads where past emails were collapsed when the thread was opened. 

I managed to fix the issue we were having by specifying an email ID for the 'msgs' in the query string (as oppose to only 'th') in the api.helper.get.email_data_pre function (see line below).

`url = window.location.origin + window.location.pathname + '?ui=2&ik=' + api.tracker.ik + '&rid=' + api.tracker.rid + '&view=cv&th=' + email_id + '&msgs=' + email_id + '&mb=0&rt=1&search=mbox';`

Hope this helps!

josteink:
Fair enough.

In that case, I think a method rename would be in order `full` makes it sounds like does more than the thing which isn't. This will probably lead many people chosing that one "just in case". Don't you agree?

Would using a identifier more in terms with the error-condition be an option, like `api.get.email_data_including_deleted`?

if you can fix that and also the `async` documentation-issue I observed, I'll be glad to merge this PR.

michi88:
Oh sorry, didn't actually see this. Doesn't this PR actually fix the whole problem? 

https://github.com/KartikTalwar/gmail.js/pull/329

josteink:
@4i60r : With [this PR](https://github.com/KartikTalwar/gmail.js/pull/329) now merged, could you check if this is still an issue?

If not, this PR probably isn't needed and should be closed.

4i60r:
@josteink It looks like it's fixed. Thanks!

josteink:
Great. Thanks for the quick feedback! I'll go ahead and close this pull request then, since it's no longer needed. 

Anyhow thank you for your contribution. patches are always welcome

PR #298, opened 2016-08-25
Fix bug: to get ID of an item with actived Preview pane, we need sear…

Navega:
Fix bug: to get ID of an item with actived Preview pane, we need search on first child of selected div on "api.dom.email_contents()".

Probably Gmail has modified its DOM...

*Navega, Software Developer at Appfluence.


danielbreves:
@KartikTalwar this doesn't seem to work anymore, reverting this commit fixed it for me (with or without preview pane).

Navega:
@danielbreves, what is the problem? My version works correctly.

PR #303, opened 2016-09-09
Corect default method

mathiasrw:
Using Google bigquery I found that your code was comparing the result of `typeof` with a variable named `undefined`.

As typeof returns a string it would compare to a string with the content `'undefined'`

As I could see you just want to have a default value of `method` i simply replaced the code with a shorter/better version...

---

This change is [Reviewable](https://reviewable.io/reviews/kartiktalwar/gmail.js/303)

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1469,7 +1469,7 @@ var Gmail_ = function(localJQuery) {
 
   api.tools.make_request = function (link, method) {
     link = decodeURIComponent(link);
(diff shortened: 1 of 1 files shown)


josteink:
Looks good to me. Consider me in support of merging this :)

KartikTalwar:
feel free to merge after conflicts have been resolved

josteink:
Considering how we were the cause of the merge conflict due to late response, consider this merge free of charge.

Thanks for the fix :)

Issue #201, opened 2015-10-29
Detect if user clicked the cancel button when sending a mail


JerryBels:
Hello !

    provider.observe.after('send_message')

allows us to detect when a mail is sent, and use this info to do... Whatever it is we do.

BUT there is an issue. This event triggers, even if the user clicks on the cancel button, thus not sending the mail ! He then proceed to change a little his email and send it again, and my program thinks he sent two.

Do you have any idea on how to solve this ?

Thanks !
Jeremy

KartikTalwar:
Hey @JerryBels would it be possible for you to clarify which cancel button is being pressed? I checked the observer by pressed the delete button on the draft but that didn't fire the event

JerryBels:
Hey @KartikTalwar my friend :)

Actually, I may have poorly explained myself. I'm talking about the moment where you send an email, immediately after clicking the "send" button Gmail gives the ability to "view the message / cancel" for a few seconds. When "Cancel" il clicked the message isn't sent.

![Like that :](http://img15.hostingpics.net/pics/78031238a.png)

Sorry for french, though ^^

So I was wondering if there was a way to detect that "cancel" event, and act accordingly. I thought about maybe detect when the "cancel" button isn't available anymore, but I wanted to know if there is a more "Gmail.js way" to do it !

Thanks

KartikTalwar:
Ah, you mean the "undo send". If you try adding the first block to L607 and the second block L657, you should be able to detect and `on('undo_send')`

Map
```js
'cs'          : 'undo_send',
```

Switch
```js
case "cs":
```

JiiBay:
That's something :) thanks ! I will try to see if it can fit in - I think yes, with minor adjusts to my API. Would you like me to update gmail.js + the doc in my fork and send a pull request ?

JerryBels:
That's something :) thanks ! I will try to see if it can fit in - I think yes, with minor adjusts to my API. Would you like me to update gmail.js + the doc in my fork and send a pull request ?

KartikTalwar:
If it works for you, a pr would be awesome!

PR #305, opened 2016-09-13
Adding the ability to detect when undo_send is clicked

JerryBels:
As we discussed here https://github.com/KartikTalwar/gmail.js/issues/201 

Pressing the button can be detected using he 

    gmail.observe.on('undo_send', function (url, body, data, xhr) {
        console.log(body, body.m);
        console.log(xhr);
    });

`body.m` holds the message id if needed.


KartikTalwar:
@JerryBels could you add the documentation on readme as well

PR #307, opened 2016-09-16
Updated make_request function to get raw massage from the new show_original page format of Gmail

bravin8:
Gmail.js API using tools.make_request function to visit the show_original page to get the email source. Since Gmail changed that page format from raw message to different html format i made some changes in the make request function to get only the raw massage.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1473,8 +1473,11 @@ var Gmail_ = function(localJQuery) {
     method  = (typeof method == undefined || typeof method == null) ? 'GET' : method;
 
     var request = $.ajax({ type: method, url: encodeURI(link), async:false });
-
-    return request.responseText;
+    var parser = new DOMParser();
+    var doc = parser.parseFromString(request.responseText, "text/html");
+    var elem = doc.getElementById("raw_message_text");
+    var source = elem.innerHTML;
+    return source;
   };
 
 


josteink:
Speaking as a fellow gmail.js user, the new format returned is indeed less useful. The RAW mime message is still there, but like you've noticed burried inside a element with id "raw_message_text".

Getting the MIME message out means re-making DOM elements, and doing lots of wasteful/inefficient operations. I'm very much for a PR like this, epsecially since it means making the api.get.email_source()-function do what it actually says in the documentation.

My only concern would be making sure this is done in way compatible across Gmail versions (since Google apps-users may be stuck on older versions for quite a bit). Is there any reliable way to do that?

@KartikTalwar: Speaking as a contributor with push-access... Would it be OK for me to approve and merge PRs like this in the future, once obvious issues like conflicts has been resolved?

KartikTalwar:
Sorry for the late replies to this pr but I agree with @josteink's comments. Compatibility becomes an issue, although I think the new message format is consistent on personal and gapps accounts. That being said, @josteink please feel free to merge any PRs that you think are good!

josteink:
Here's my proposal:

1. Wrap the Domparser logic in a try-catch.
2. Check if you can find the named element.
3. If any error happens, or the element can not be found... Then assume response-text is the RAW mime-message and that any further processing is unneeded.
4. Otherwise return contents of element, as your code already does.

I think that should address all compatibility-concerns, and I'd be happy to merge that.

josteink:
I'm eager to get old PR's closed, so I took some liberties:

* I took the existing implementation, and made the changes I suggested.
* I moved the *location* of the code to email_source (where it logically belongs) instead of in make_request.

With that, `api.get.email_source()` should now always return MIME-messages, as the documentation says it does.

Thanks for the patch!

bravin8:
@josteink Great work mate.

PR #311, opened 2016-09-23
Fixing mail_id parsing + adding module.exports if possible

nedarb:
Support module.exports when present, otherwise set the window.Gmail instance.

Also add missing semicolons (jshint warnings)

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -203,7 +203,7 @@ var Gmail_ = function(localJQuery) {
     }
 
     return chat[0].getAttribute('aria-labelledby') == ':wf';
-  }
+  };
 
   api.check.should_compose_fullscreen = function(){
     var bx_scfs = [];
@@ -213,7 +213,7 @@ var Gmail_ = function(localJQuery) {
       bx_scfs = ['bx_scfs','false'];
     }
      return (bx_scfs[1] == 'true' ) ? true : false;
-  }
+  };
 
 
   api.check.is_google_apps_user =function() {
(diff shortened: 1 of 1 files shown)


nedarb:
@KartikTalwar here's the updated PR

josteink:
The best way to get a PR accepted is usually to make it constrained to only fixing *one* thing.

And in that regard, you can leave out making things Node-friendly, because that was fixed seperately in https://github.com/KartikTalwar/gmail.js/pull/323 which is out on npm now :)

src/gmail.js
```diff
@@ -213,22 +213,25 @@ var Gmail_ = function(localJQuery) {
...
 
 
   api.check.is_google_apps_user =function() {
     var email = api.get.user_email();
     return email.indexOf('gmail.com', email.length - 'gmail.com'.length) == -1;
   };
 
+  api.check.is_in_popout = function() {
+    return api.tools.parse_url(window.location.href).view === 'btop';
+  };
```
josteink: This method should also be added to the documentation (and if possible also to the TypeScript type declarations).

src/gmail.js
```diff
@@ -370,7 +373,8 @@ var Gmail_ = function(localJQuery) {
       }
     }
     else {
-      hash = api.tools.parse_url(window.location.href).th;
+      hash = api.tools.parse_url(window.location.href).th || 
+        window.location.hash.split("/").pop().replace(/#/, '').split('?')[0];
     }
```
josteink: Again, some comments or context on what's getting fixed here would be helpful.

josteink:
@nedarb: I'm very much in support for fixing jshint/eslint errors. Appreciate the effort.

That said: I've created a separate issue for doing it 100% here: https://github.com/KartikTalwar/gmail.js/issues/327

Feel free to chime in with your opinion there :)

With the NodeJS part already merged separately, what remains here is then a few unspecified fixes, which is very hard to have an opinion on without further context.

If you provide that, I can consider merging this, but otherwise I think it's better to close it.

josteink:
Both issues mentioned in PR subject has been fixed in other, merged PRs. Closing this one.

PR #321, opened 2016-10-22
Revert "Fix bug: to get ID of an item with actived Preview pane, we need sear…"

danielbreves:
Reverts KartikTalwar/gmail.js#298

Tested with and without preview pane mode.

screen shot 2016-10-23 at 10 47 02 am

cc @KartikTalwar, @Navega

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -354,7 +354,7 @@ var Gmail_ = function(localJQuery) {
         var text = [];
 
         for(var i=0; i<items.length; i++) {
-          var mail_id = items[i].children[0].getAttribute('class').split(' ')[2];
+          var mail_id = items[i].getAttribute('class').split(' ')[2];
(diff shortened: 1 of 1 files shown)


josteink:
Gmail often has different styling/CSS and other things UI-wise depending on your edition, your Google Apps account etc...

I think a safer/more conservative approach might be checking for both values?

josteink:
Another PR which addresses some of these issues, but which is a lot less atomic and less focused can be found here: https://github.com/KartikTalwar/gmail.js/pull/311

It has a fix for the same issue which seems to handle both ways:
https://github.com/KartikTalwar/gmail.js/pull/311/files#diff-ed1a8db78264acfe07214717bdad5425R357

If you can instead use that, and check that everything still works, I'll get this merged.

josteink:
You know what? I'll fix that up myself. Thanks for the patch :)

PR #322, opened 2016-11-04
Add basic TypeScript support.

josteink:
TypeScript allows you to write JavaScript, but with strong type-checking. For big and complex APIs like Gmail.js this ensures all objects, functions and returned values are used correctly, not to mention having a reliable and correct in-editor auto-completion.

According to [Definitely Typed](https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package) the recommended way of adding support to existing libraries is to send a PR to the library itself.

So here it is: This PR contains a type-declarations for the existing gmail.js code.

This enables TypeScript users to use gmail.js in their project and maintain type-safety by simply referencing the provided `gmail.d.ts`-file.

As long as the gmail.js API does not change, no changes will need to be made to the gmail.d.ts file either.

To develop and test the type declarations-file you will need to do the following things:

````sh
# get npm dependencies
npm install
# get typescript jquery type-declarations file
./node_modules/.bin/typings install
# compile and verify gmail.js type-declarations
./node_modules/.bin/tsc
````

I think this PR would be a nice addition to gmail.js and really hope you will consider merging it.


josteink:
Typings has been obsoleted for npm @types.

I'm closing this PR and will re-open another one with more up-to-date code.

PR #324, opened 2016-11-10
Add basic TypeScript support.

josteink:
TypeScript allows you to write JavaScript, but with strong type-checking. For big and complex APIs like Gmail.js this ensures all objects, functions and returned values are used correctly.

This PR contains a type-declarations for the existing gmail.js code.

This enables TypeScript users to use gmail.js in their project and maintain type-safety by simply referencing the provided gmail.d.ts-file.

As long as the gmail.js API does not change, no changes will need to be made to the gmail.d.ts file either.

To develop and test the type declarations-file you will need to do the following things:

````sh
npm install
./node_modules/.bin/tsc
````

I think this PR would be a nice addition to gmail.js and really hope you will consider merging it.


KartikTalwar:
@josteink thanks for creating this PR. I don't have any problems with having a typescript declaration but the issue is that I and others submitting PRs would have to ensure that the new changes are made to the ts file as well. You already pointed this out in your PR and given that the api already got a new function a few minutes ago (via another pr), I'd be willing to merge this if you'd be willing to take the responsibility for ensure that the ts file is up to date as well.

josteink:
Sure. I'll do my best. 

The worst thing which can happen though is that new APIs are provided a little bit slower to the typescript crowd, and I don't think that's going to be a big problem. 

For now I'll set to watch this repo to keep up to date and update my PR according to the new functions available.

I'll let you know once those commits are in place and we'll take it from there.

KartikTalwar:
Sounds good, thanks

josteink:
As far as I can see, that should cover all the new changes. Merge at will :)

PR #326, opened 2016-11-12
Updated unread_inbox_emails to parse properly when Gmail has the "Unread items in the first section and inbox, for example (2 : 3)" option set

Thomas101:
When using priority inbox the unread count in the gmail UI can be setup to display two counts in the format `(10 : 20)`. At present this parses as `1020` rather than `10`.

Example on gmail...
screen shot 2016-11-12 at 08 30 39

Settings to configure this behaviour...
5675335c-a66b-11e6-8eea-abeaea9567ee

This pull request splits the count by colon first and takes the first value to return the correct value


KartikTalwar:
Thanks

Thomas101:
Wow merging at the speed of light ;-). Thanks!

PR #329, opened 2016-11-17
fix getting the original source from email

michi88:
This fixes that we really get the original source of an email including attachments etc.

The current implementation only includes a truncated version.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -379,7 +379,7 @@ var Gmail_ = function(localJQuery) {
     else {
       hash = api.tools.parse_url(window.location.href).th;
     }
-    
+
     return hash;
   };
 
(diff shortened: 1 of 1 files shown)


josteink:
I haven't experienced truncated email attachments myself, but I'll take your word for it.

While I like that approach much better than the existing one *in principle*...

One issue I'm getting with this (in an extension-context) is that I'm being blocked by CORS when trying to fetch it, since (for whatever reason) it seems to be redirected to an external domain which is not mail.google.com:

https://mail-attachment.googleusercontent.com/attachment/u/0/?view=att&th=1586c9ed29dfe191&....

In what context are you able to use this patch without triggering CORS warnings?

josteink:
Strange. Now it's working for me again. OK for me I guess.

@KartikTalwar: You OK with this patch too? I'd love to have more than one tester when we change stuff like this :)

josteink:
You know what? I can't see how this fix could possibly lead to other issues, given how local this fix is.

Consider it merged. Thanks for you contribution!

michi88:
Thanks for being a collaborator!

michi88:
@josteink This was really strange!! I see the CORS errors now also! But not always... depends on if the browser cached the first GET call. If the browser used the cached redirect to https://mail-attachment.googleuserconten.... google doesn't include the Access-Control-Allow-Origin header and thus the preflight check fails.

I'm now busting the cache and that works. PR coming!

PR #333, opened 2016-11-22
Email source CORS fix

michi88:
As already noticed by @josteink in the merged PR https://github.com/KartikTalwar/gmail.js/pull/329 there is a CORS problem when the browser caches the first redirect. 

Although it seems that the location of the redirect is the same when requested 'fresh' but Google only supplies the `Access-Control-Allow-Origin` header for the latter. Strangely!!! Btw, this might also happen in other places so if you see CORS related errors check that the cache is not used. And of course to learn from this... do not disable cache in devtools when testing gmail.js! ;)

jQuery's `{cache: false}` adds something like this to the url `_=76348726487` which might not be OK for all GET calls so I've added the flag on `make_request` and `make_request_async`.

I've also added a change to `is_inside_email` because when email text is hidden (collapsed because of thread already read) it should return `true`.
This happens for example when someone emails an attachment as a reply on an email but without adding text.

This fixes an issue in my application and I don't see any problems but if you don't want to add it I'll make you a new PR if you want.

The `TODO` probably needs to looked at because I see some errors (not a lot) coming in from my extension in production.

### DIFF
--- a/src/gmail.d.ts
+++ b/src/gmail.d.ts
@@ -228,9 +228,9 @@ interface GmailGet {
     */
     email_source(email_id: string): string;
     /**
-       Does the same as email_source but accepts a callback function
+       Does the same as email_source but accepts a callback and an optional error_callback function
     */
-    email_source_async(email_id: string, callback: (email_source: string) => void): void;
+    email_source_async(email_id: string, callback: (email_source: string) => void, error_callback: (jqxhr, textStatus: string, errorThrown: string) => void): void;
     displayed_email_data(): GmailEmailData;
 
 }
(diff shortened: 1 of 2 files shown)


josteink:
A few comments :

*  please remove the todo comment and register a official github issue instead so that it can be tracked properly. 
*  please submit that is inside email pics in the separate PR so that it can be discussed separately. 
*   Looking at the code it seems like users of the Gmail.js library will have to manually implement cache bursting themselves by using the error callback. Is this correct? Wouldn't it be better to implement the once and properly inside the library ?

josteink:
A few comments :

* It's awesome that you're updating the TypeScript bindings. That's really appreciated :)
* Please remove the todo comment and register a official github issue instead so that it can be tracked properly, and is more easily discoverable for other people having the same issue.
* Please submit the "is inside email" bits in a separate PR so that it can be discussed separately.

Back to the main gist of this PR: Looking at the code it seems like users of the Gmail.js library will have to manually implement cache bursting themselves by using the error callback.

Is this correct? Wouldn't it be better to implement the once and properly inside the library? Or am I wrong and this PR fully fixes the CORS issue without any further intervention needed?

michi88:
No intervention is needed as `email_source_async` calls `make_request_async` with `disable_cache=True`.
I agree `make_request_async` might not be the nicest api but as it's public we can't change the first 3 arguments. Do you have another suggestion? 

Will make a separate PR and make an issue.

josteink:
As long as that works I'm perfectly happy with the code as it is. No need for other changes IMO. 

Just clean up the other aspects I asked for and I'll get this merged. :)

michi88:
new PR here https://github.com/KartikTalwar/gmail.js/pull/335

PR #334, opened 2016-11-22
fix is_inside_email when email text is hidden/collapsed

michi88:
Fix for `is_inside_email` because when email text is hidden (collapsed because of thread already read) it should return true.

This happens for example when someone emails an attachment as a reply on an email but without adding text.

I don't see when we might want to check for that visibility but it probably did have a reason when it was added (a long time ago btw).

Ideas anyone?

@josteink as discussed in https://github.com/KartikTalwar/gmail.js/pull/333

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -302,9 +302,7 @@ var Gmail_ = function(localJQuery) {
         for(var i=0; i<items.length; i++) {
             var mail_id = items[i].getAttribute("class").split(" ")[2];
             if(mail_id !== "undefined" && mail_id !== undefined) {
-                if($(items[i]).is(":visible")) {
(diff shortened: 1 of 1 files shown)


josteink:
I honestly don't know.

Would you be able to dig up who originally wrote this line of code via git blame?

That would probably help us find out why the code is the way this.

michi88:
https://github.com/KartikTalwar/gmail.js/commit/8e395180c81d2de024817e4d1d3128a4b02e77a9

Ping @KartikTalwar

KartikTalwar:
Definitely had a reason in 2013 (was required for an extension I was building) but thanks for noticing this - it's no longer relevant.

PR #335, opened 2016-11-22
Email source cors cache

michi88:
New PR as discussed in https://github.com/KartikTalwar/gmail.js/pull/333

---

As already noticed by @josteink in the merged PR #329 there is a CORS problem when the browser caches the first redirect.

Although it seems that the location of the redirect is the same when requested 'fresh' but Google only supplies the `Access-Control-Allow-Origin` header for the latter. Strangely!!! Btw, this might also happen in other places so if you see CORS related errors check that the cache is not used. And of course to learn from this... do not disable cache in devtools when testing gmail.js! ;)

jQuery's `{cache: false}` adds something like this to the url `_=76348726487` which might not be OK for all GET calls so I've added the flag on `make_request` and `make_request_async`.

### DIFF
--- a/src/gmail.d.ts
+++ b/src/gmail.d.ts
@@ -228,9 +228,9 @@ interface GmailGet {
     */
     email_source(email_id: string): string;
     /**
-       Does the same as email_source but accepts a callback function
+       Does the same as email_source but accepts a callback and an optional error_callback function
     */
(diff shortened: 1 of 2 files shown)


josteink:
LGTM.  Thanks for contributing!

PR #337, opened 2016-11-23
start parameter calculation change at visible_emails_pre function

avzabr:
The markup for email's pagination has been changed at gmail recently:

image

Was:
`<span class="Dj"><b>111</b>–<b>114</b> of <b>114</b></span>`
New:
`<span class="Dj"><span class="ts">111</span>–<span class="ts">114</span> of <span class="ts">114</span></span>`

Also if there are more than 1000 emails the number may be "1,234". The ` .replace(",", "")` has been added to fix it. 

Note: i don't know why we need duplicated start parameter.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1591,9 +1591,11 @@ var Gmail_ = function(localJQuery) {
     api.helper.get.visible_emails_pre = function() {
         var page = api.get.current_page();
         var url = window.location.origin + window.location.pathname + "?ui=2&ik=" + api.tracker.ik+"&rid=" + api.tracker.rid + "&view=tl&num=120&rt=1";
-        if ($(".Dj:visible").find("b:first").text()) {
-            url += "&start=" + parseInt($(".Dj:visible").find("b:first").text() - 1) +
-                "&start=" + parseInt($(".Dj:visible").find("b:first").text() - 1);
+        var start = $(".aqK:visible .Dj").find("span:first").text().replace(",", "");
+        if (start) {
+            start = parseInt(start - 1);
+            url += "&start=" + start +
+                   "&start=" + start;
         } else {
             url += "&start=0";
         }
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1595,7 +1595,7 @@ var Gmail_ = function(localJQuery) {
         if (start) {
             start = parseInt(start - 1);
             url += "&start=" + start +
-                   "&start=" + start;
+                   "&sstart=" + start;
         } else {
             url += "&start=0";
         }


josteink:
Thanks for the patch! I've done some rudimentary checking and can confirm that:

* the existing code *doesn't* detect page-number.
* the new code *does*.

And by those standards, this looks good to me.

> Note: i don't know why we need duplicated start parameter.

This was introduced in the following commit: a5c9928d152ccb7416725ab672c0991f33376be6

As far as I can tell, this is actually a code-error. See the previous state: It only used one start-parameter. Could you test if this works as it should if you don't setup duplicate starts?

If so, it would be nice to clean that up at the same time.

@avzabr: Any comment on this?

avzabr:
@josteink 
Good finding! Actually based on the mentioned commit: [a5c9928](https://github.com/KartikTalwar/gmail.js/commit/a5c9928d152ccb7416725ab672c0991f33376be6) the second parameter should be the `sstart` instead of duplicated `start`. 

image

I renamed the second parameter to `sstart` with the b4e1e76 commit

josteink:
Unsure about this one. That `sstart` was supposedly a typo as can be seen in this follow-up commit: 0e9822088ce10ddfbcbd383181793bbb5bba811d by Lucas Shculze (Github ID unknown).

However, the original code, before @Navega and Lucas made their changes only had one single `start`-value. I honestly think the second `sstart` (or `start` in the typo-fixed version) is a mistake. I don't think it should be there at all.

What I tried to ask you (but obviously failed!), was if you would be able to test without the second `start` entry at all, and see if that works. Could you do that please? :)

Because, if it's a bug, we should clean this up at the same time IMO.

@Navega: Do you have any opinion on this?

avzabr:
@josteink Ok, i was able to do several tests.

Test procedure:
1. Download [gmail-chrome-extension-boilerplate](https://github.com/KartikTalwar/gmail-chrome-extension-boilerplate)
2. Update the gmail.js with new version from pull request 
3. Update the `num` parameter to `10`
4. Update the gmail inbox settings to show 10 emails (as i had only 70 emails in my mailbox)
5. Ensure the gmail.js with `sstart` parameter 
6. Get `copy(gmail.get.visible_emails())` from Primary tab first/last pages and Social tab first/last pages
7. Remove the `sstart` parameter from gmail.js
8. Get `copy(gmail.get.visible_emails())` from same pages
9. Compare results from 6) and 8) with jsondiff

Test results are attached: [GmailJsPull337.zip](https://github.com/KartikTalwar/gmail.js/files/610796/GmailJsPull337.zip)

The result is following:
1) At Primary first page, Social tab first/last pages results are same
2) At Primary last page the json received without `sstart` parameter does not correspond to the emails displayed on a screen

Thus from the test results the sstart parameter is needed.

josteink:
That's great testing! Thanks for putting in the effort! (y)

If it `sstart` is indeed needed, then we keep it. Simple as that.

Consider this merged. Thank you for your effort! :)

Navega:
Hi guys,

I have been debugging the problem. Gmail modifies its structure continuously. My commit does not work because Gmail replaces label "<b>" with "<span>" where number of offset page is positionated. I used 2 parameters: "start" and "sstart" because they depend of the used version. Older versions used the parameter **"start"**, but now the correct parameter is **"sstart"** to ask mails with offset.

Certainly. good catch with offset > 1.000, thanks!

Do you agree?

Navega:
@josteink The changes are perfect, thanks. The solution is to replace find("b:first") by find("span:first").

josteink:
Thanks for chiming in about the history of these parameters. That makes things much more clearer :)

And just to be clear: In your opinion, no more changes needs to be done or reverted, right?

Navega:
Yes, only a change: Spanish version uses "." in place of "," to write the offset. Coul you add this?
- var start = $(".aqK:visible .Dj").find("span:first").text().replace(",", "");
+ var start = $(".aqK:visible .Dj").find("span:first").text().replace(",", "").replace(".","");

josteink:
Sounds fair enough. I'll get it done.

PR #338, opened 2016-11-24
is_deleted sometimes is not calculated correctly

avzabr:
When the email is deleted the x[13] is returned null from sever:
image
Thus the comparison `(x[13] === underfined) ?` is not enough for this case. And due to this there is a scenario when the gmail.get.displayed_email_data() throws an error: 

1. In a email thread remove email
2. Execute gmail.get.displayed_email_data() call
Error: 
`gmail.js:1862 Uncaught TypeError: Cannot read property '6' of null(…)
Gmail_.api.tools.parse_email_data 
@ gmail.js:1862Gmail_.api.helper.get.email_data_post 
@ gmail.js:1906Gmail_.api.get.email_data 
@ gmail.js:1915Gmail_.api.get.displayed_email_data 
@ gmail.js:1969(anonymous function) 
@ VM1028:1`

image

To fix that the the `api.tools.parse_email_data` method has been updated: `(x[13] === underfined) ?` replaced with comparison with `x[13] ?`

Also the `data.threads[x[1]].labels = x[9];` has been added to be able to retrieve the labels for the email.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1851,7 +1851,7 @@ var Gmail_ = function(localJQuery) {
                 }
 
                 data.threads[x[1]] = {};
-                data.threads[x[1]].is_deleted = x[13] === undefined;
+                data.threads[x[1]].is_deleted = x[13] ? true : false;
                 data.threads[x[1]].reply_to_id = x[2];
                 data.threads[x[1]].from = x[5];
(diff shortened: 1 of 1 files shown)


josteink:
Can confirm error. Fix looks good to me.

Thanks for another nice patch :)

PR #339, opened 2016-11-24
Added api.get.displayed_email_data_async, updated is_deleted calculation

avzabr:
It has been noticed that if the `x[13] ` is `null` it does not mean the message is deleted. For e.g. the `x[13]` may be `null` if it is collapsed or the message is plaintext and very small. 

But as far i can see if if the message contains the `"^k"` label it means that the message is actually deleted.
image

Thus the change is to set `is_deleted` flag based on the existence of "^k" label. 
Also the `api.get.displayed_email_data_async` has been added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 
 # Changelog
 
+## Version 0.6.4
+
+- Fix `api.get.displayed_email_data` when email text is deleted
+
 ## Version 0.6.3
 
 - Fix `api.check.is_inside_email` when email text is hidden/collapsed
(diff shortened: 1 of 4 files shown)


josteink:
You're adding a new function. Which is welcome :)

But with new functions comes responsibility... :)

* Do you know how to add a typescript type-declaration for this new function to `gmail.d.ts`? If you could add that, it would be nice.
* Function should also be added to documentation on `README.md`.
* Adding it to `CHANGELOG.md` too would be nice.

Sounds doable?

avzabr:
Yep, no probs.

avzabr:
Closing as the code issue has been found.

PR #340, opened 2016-11-24
Displayed email data async

avzabr:
It has been noticed that if the `x[13] ` is `null` it does not mean the message is deleted. For e.g. the `x[13]` may be `null` if it is collapsed or the message is plaintext and very small. 

But as far i can see if if the message contains the `"^k"` label it means that the message is actually deleted.
image

Thus the change is to set `is_deleted` flag based on the existence of "^k" label. 
Also the `api.get.displayed_email_data_async` has been added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix `api.check.is_inside_email` when email text is hidden/collapsed
 - Fix `api.get.visible_emails`
 - Fix `api.tools.parse_email_data` when email text is deleted
+- Fix `api.tools.get_reply_to` when input parameter is null
 - Added `api.get.displayed_email_data_async`
 
 ## Version 0.6.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 
 # Changelog
 
+## Version 0.6.4
+
+- Fix `api.tools.parse_email_data` when email text is deleted
+- Added `api.get.displayed_email_data_async`
+
 ## Version 0.6.3
 
 - Fix `api.check.is_inside_email` when email text is hidden/collapsed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,12 @@
 
 # Changelog
 
-## Version 0.6.4
-
-- Fix `api.tools.parse_email_data` when email text is deleted
-- Added `api.get.displayed_email_data_async`
-
 ## Version 0.6.3
 
 - Fix `api.check.is_inside_email` when email text is hidden/collapsed
 - Fix `api.get.visible_emails`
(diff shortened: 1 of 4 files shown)


josteink:
This fix also seems a bit incomplete. i tried going to the trash and run `gmail.get.displayed_email_data()`. This failed with a similar failure mode a different place in the codebase:

https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js#L1822

Basically `ms13` here is `null` too, not `undefined`. Can you reproduce this on your gmail account? If so we should apply the same shorthand fix here.

I think that before we close this issue, we should fix it properly, by testing all possible places this function could be called. And clearly we still have some way to go.

On a sidenote, I'd also appreciate it if we could keep each PR constrained to adressing one single issue. It makes things much simpler to work with.

If you want to fix several things at once, use several branches and send several PRs :)

So in short:

* We need tpl ug the `ms13` thing. Are there other things we need to plug?
* Once we have the actual error fixed, I *then* think adding new APIs are appropriate (or you can do it in a seperate PR).
* You can also leave the changelog items at version 0.6.3. It's not released yet :)

josteink:
To dig further into this one... In `api.tools.parse_data()` when it calls into `api.tools.get_reply_to(x[13])`...

On un-deleted emails this works fine. But on deleted emails, (with my fix) I get a blank list of `people_involved` which clearly isn't right.

This at the same time as we have the sender (of the deleted email) found on `x[4]`, `x[5]` and `x[6]`. So the information *is* available. Just elsewhere. In another format.

In general, it seems that we cannot reliably assume that all normal information is as available on deleted emails as it is on normal emails.

Is this a problem? Is this something that we're willing to complicate the code we have to work with in order to solve?

@KartikTalwar: Your opinion is appreciated.

avzabr:
@josteink 

> Basically ms13 here is null too, not undefined. Can you reproduce this on your gmail account? If so we should apply the same shorthand fix here.

I was able to reproduce the error here and applied fix to `api.tools.get_reply_to` function. But at the same time with this fix i can see the `people_involved` for the email in trash. May be there are different type of emails? Draft/Sent/Inbox? 
image

> So in short:
> 
> 1. We need to plug the ms13 thing. Are there other things we need to plug?
> 2. Once we have the actual error fixed, I then think adding new APIs are appropriate (or you can do it in a > seperate PR).
> 3. You can also leave the changelog items at version 0.6.3. It's not released yet :)

The first and third points are fixed. And i agree it would be better to separate these changes in two pull requests:

1. Addition of new api method `gmail.get.displayed_email_data()`
2. `is_deleted` calculation based on `^k` label

May we leave it as is this time?

josteink:
Fair enough. Consider this merged, but please try to submit separate PRs for separate features next time :)

Issue #336, opened 2016-11-22
Window.opener === null when registering xhr_watcher


michi88:
I've see some (not a lot) errors coming in from my extension in production.

Happens when `api.tools.xhr_watcher` is called and this line is executed:

`var win = top.document.getElementById("js_frame") ? top.document.getElementById("js_frame").contentDocument.defaultView : window.opener.top.document.getElementById("js_frame").contentDocument.defaultView;`

So the question really is in which cases don't we have `js_frame` and what exactly do we want to do in that case?

josteink:
I think I would try to decompose that section into its own function interspersed over several lines of code. 

That way you could put a breakpoint on the line where window.opener is probed (and fails!) and from that breakpoint you should be able to inspect the conditions under which this error occurs. 

That would probably be a good start  before we decide on which changes we need to make. 

Sounds like a plan?

michi88:
Yeah, so this is the error I found:

![image](https://cloud.githubusercontent.com/assets/1010260/20531525/571cf7b0-b0d7-11e6-9b00-a7d7d62ab772.png)

As the url is `https://mail.google.com/` and not `https://mail.google.com/mail/u/0/` maybe I'm calling the watcher too soon?

This is what I do currently:

```javascript
var initSRforGmail = function initSRforGmail() {

  var gmail;

  var refresh = function refresh(f) {
    if (/in/.test(document.readyState) || typeof Gmail === 'undefined' || typeof window.jQuery === 'undefined') {
      setTimeout(function () {
        refresh(f);
      }, 10);
    } else {
      f();
    }
  };

  var main = function main() {

    gmail = new Gmail();
    gmail.observe.on('open_email', function () {
      someFunc();
    });
    // .... other stuff including more gmail.observe.on calls
  }
  refresh(main);
};

try {
  initSRforGmail();
} catch (e) {
  SRErrorHandler.captureException(e);
}
```

michi88:
Oh and by the way... I can't reproduce it :)

I don't see this error often and I have nobody complaining that it doesn't work (no big deal) but we might have some logic here that needs to be changed.

michi88:
Ah maybe I should use `gmail.observe.on("load")` first before I register the other watchers as gmail needs to have loaded the `top.document.getElementById("js_frame")` iframe. 

```javascript
        else if(action === "load") {

            // wait until the gmail interface has finished loading and then
            // execute the passed handler. If interface is already loaded,
            // then will just execute callback
[...]

PR #341, opened 2016-11-24
throw error when the js_frame hasn't been inserted yet

michi88:
as discussed in https://github.com/KartikTalwar/gmail.js/issues/336


josteink:
That looks great. Again: Thanks!

PR #359, opened 2016-12-21
Stop being a singleton and monkey patch XMLHttpRequest in a chain ins…

patrickriordan:
…tead

Was running a modified version of GmailJS that did run as a singleton and none of it's xhr_watchers got called because of the way we monkey patch. If we monkey patch in a chain and don't expose the original XMLHttpRequest then there is no reason multiple gmailJSes can't run in one gmail.

Also being a singleton is bad idea since different versions will battle each other.

This reverts #284 and instead monkey patches in a chain.

@josteink This is what I was running into before, sorry for being confusing.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -4,7 +4,7 @@
 // https://github.com/KartikTalwar/gmail.js
 //
 
-var Gmail_ = function(localJQuery) {
+var Gmail = function(localJQuery) {
 
     /*
       Use the provided "jQuery" if possible, in order to avoid conflicts with
@@ -945,8 +945,11 @@ var Gmail_ = function(localJQuery) {
     };
 
     api.tools.parse_requests = function(params, xhr) {
-        params.url_raw = params.url;
-        params.url = api.tools.parse_url(params.url);
+        // We should not be modifiying the original object
+        // but because we used to and still do we need to
+        // handle it incase an old version of gmailjs is running
+        params.url_raw = params.url_raw || params.url;
+        params.url = api.tools.parse_url(params.url_raw);
         if(typeof params.body === "object") {
             params.body_params = params.body;
             params.body_is_object = true;
@@ -991,12 +994,10 @@ var Gmail_ = function(localJQuery) {
             }
             var win = js_frame.contentDocument.defaultView;
 
-            if (!win.gjs_XMLHttpRequest_open) {
-                win.gjs_XMLHttpRequest_open = win.XMLHttpRequest.prototype.open;
-            }
+            api.tracker.xhr_open = win.XMLHttpRequest.prototype.open;
 
             win.XMLHttpRequest.prototype.open = function (method, url, async, user, password) {
-                var out = win.gjs_XMLHttpRequest_open.apply(this, arguments);
+                var out = api.tracker.xhr_open.apply(this, arguments);
                 this.xhrParams = {
                     method: method.toString(),
                     url: url.toString()
@@ -1004,9 +1005,7 @@ var Gmail_ = function(localJQuery) {
                 return out;
             };
 
-            if (!win.gjs_XMLHttpRequest_send) {
-                win.gjs_XMLHttpRequest_send = win.XMLHttpRequest.prototype.send;
-            }
+            api.tracker.xhr_send = win.XMLHttpRequest.prototype.send;
 
             win.XMLHttpRequest.prototype.send = function (body) {
                 // parse the xhr request to determine if any events should be triggered
@@ -1040,7 +1039,7 @@ var Gmail_ = function(localJQuery) {
                 }
 
                 // send the original request
-                var out = win.gjs_XMLHttpRequest_send.apply(this, arguments);
+                var out = api.trakcer.xhr_send.apply(this, arguments);
 
                 // fire on events
                 api.observe.trigger("on", events, this);
@@ -2799,19 +2798,8 @@ var Gmail_ = function(localJQuery) {
     return api;
 };
 
-function initializeOnce(fn) {
-    var result;
-    return function() {
-        if (fn) {
-            result = fn.apply(this, arguments);
-        }
-        fn = null;
-        return result;
-    };
-}
-
 // required to avoid error in NodeJS.
-var GmailClass = initializeOnce(Gmail_);
+var GmailClass = Gmail();
 if (typeof(window) !== "undefined" && !window.Gmail) {
     window.Gmail = GmailClass;
 }


josteink:
Hey patrick!

Thanks for the work and thanks for the patch!

I agree chained monkey-patching sounds like the only way to go to ensure correct operation when working with other extensions.

And obviously, having a singleton implementation is going to cause issues there, so thanks for looking into that as well. Myself, I never quite knew the history behind the singleton-implementation, so kudos for digging that up in the process!

With that said, there's a few issues here:

* The CI build is broken. I'm not merging any patch which breaks the CI-build. So next time please look into the log and see what's failing, and check if it's something you think you can fix. If you can't, feel free to page me about it.

* The build actually fails because NodeJS module-export is broken. I'll provide some inline comments on how you can go about fixing that.

* I don't know the full history behind the original patch. I can take a basic look by manually creating 2 instances and hooking up some observers and see that everything still works out, but I would be even more comfortable if @KartikTalwar who made the original patch and possibly @derekribbons who first reported these bugs would be able to look into this too. (Feel free guys!)

With that said, if everything else clears I see no issue with this patch.

josteink:
Any news here?

patrickriordan:
@josteink Haven't had a chance to work on this because I've been on vacation. I will try to fix those issues this week. Thanks for taking a look!

derekribbons:
Hi All,

Garris submitted the patch because of the way gmail.js replaces the native
window.Gmail can only be done once.  Currently gmail.js is being used in a
number of Chrome plugins (LinkedIn Sales Navigator, SalesforceIQ and the
folks @ InboxSDK + more).  If gmail.js clobbers the main window object more
than once, Gmail dies.

Happy to have the singleton go away, but the other issue will have to be
resolved.  Garris and I are on the LinkedIn team and we have contacts on
the IQ team, so happy to test other solutions, but pretty much anyone that
has a Gmail add-on will be impacted if it is not resolved.

-Derek

josteink:
Hey Derek.

Thanks for chipping in. That's exactly the kind of feedback I was looking for!

So from what I can tell, what you're basically saying is that if we're doing any changes to the singleton structure, we should absolutely test with -two- different extensions built from the new code, and have them both installed at the same time, while checking that Gmail still works? Am I correct in that?

If so, that is certainly doable, but was definitely beyond my initial plans. To me it also sounds like a booby-trap waiting to go off... In theory it sounds like something where investing in more automated testing would be good, but I'm not sure how I would go about doing that. Any opinions?

derekribbons:
Hi Jostein,

Correct.

We could reproduce the issue in the past by just loading a second instance
of gmail.js, we would inject content into a DIV that included gmail.js,
then we could inject the same thing again into that DIV and start clicking
around Gmail and get the failure in seconds.  We preload our libraries once
when the plugin loads, so it was never really an issue in early testing,
but when there are 2 plugins doing the same thing, that is when we had to
update gmail.js.  We reached out to the IQ team once Garris had a fix as
they were who we were doing our compatibility testing against, they took
the patch once it was pushed to master.

In your case, you could probably just load up SalesforceIQ and then have a
stub that loads gmail.js again and it should generate the dreaded
"Something went wrong...." in Gmail (provided the singleton mode is
disabled).  Basically when window.Gmail gets replaced the second time, that
is when the issue occurs.

Hope this helps,
-Derek

On Tue, Jan 3, 2017 at 11:26 AM, Jostein Kjønigsen <notifications@github.com

garris:
Hi guys.  I concur with everything @derekribbons said.  Here's some detail...

We went the singleton route because at the time it was the most straightforward way to solve the multi-instance issues.  It required very few changes so we had pretty good confidence that it would be backwards compatible for vanilla implementations.  Which turned out to be correct -- except in your case where I gather you're doing something outside the API?

Also,  for development, I built a test harness in node which which appends a console script and tests multiple injection cases -- I was considering publishing that but it would have required a lot of documentation and just didn't seem worth it.   I'd recommend something automated if you go ahead with a refactor.  You'll likely have too much uncertainty if you attempt to refactor manually (by clicking around).   

You would also want to make sure it doesn't break older implementations e.g. So current version and new version will play nice -- there is a non-trivial installed base out there.   Perhaps you might consider forking if your changes are significant enough.

josteink:
@patrickriordan: I'd love to just go ahead and move all this code to ES5 and use proper classes and all that nonsense. Maybe even make it workable within TypeScript. And for that, such a cleanup like this would certainly not only help, but be absolutely necessary.

But given the feedback so far, it seems the actual amount of work which needs to be done complete this change is non-trivial. And it's at that point where you have to chose between "nice to have" and "must have".

I'd rather consider keeping this working as is ("must have"), rather than risk major, breaking changes over something which would merely be "nice to have". Especially if we also risk alienating a community, or maybe even causing a fork.

Would it be possible for you to try to investigate other means of solving your issue than this current approach?

josteink:
Sure. As long as these other gentlemen are happy too, I'm perfectly OK with that.

josteink:
> My issue happens because my version does not run as a singleton. Any version that runs as a singleton can be clobbered by an older singleton version and break the app anyways.

On a sidenote: What do you mean with "my version"? Are you using a forked, non-official version?

If so, there's not much we can do here. Are there any particular reasons you are running your own version?

josteink:
So to review this issue once again (and thanks to everyone involved!):

* We **can not** patch XHR things in a chain. Because for some reason it breaks stuff in unexpected ways.
* So if we want to stop having the GmailJS-class being a Singleton, we need to extract the XHR-patching out to a singleton XHR-patcher, which we can use from our non-Singleton GmailJS-instance.
* If we can do that, we can rewrite GmailJS to use a real class, use real ES6 constructs and in general move more in a ES6ish direction in the future. Which sounds like a very good thing.

As far as I can tell, the whole feasibility of this "project" depends on our ability to extract the XHR intercepting to a singleton. We can see the following in the [relevant code](https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js#L985-L1050):

* We create a "global" variable of the original XHR `open` and `send` functions (which is good, it enables detection)
* We rewrite the functions using class-internal closure-functions.
* The good news seems to be that these class-internal closures only seem to reference world-accessible class instance-members.

But there are lots of places where it (IMO) gets tricky:

* If we were to extract this to a separate singleton class... That singleton class would need to know about *all* the GmailJS-instances it was to serve. That is: GmailJS-instances would need to register with it, to be called back into.
* The "am I initialized"-flag is instance-local. Older versions of GmailJS would observe that the original handles were already saved, but overwrite the global Singleton XHR interceptor's handlers. We might be able to work around this doing hacky stuff with `setInterval`, but do we really want to?

Basically: How can we best handle the co-existance of "old" GmailJS code and new one at the same time? There are several aspects which might be hard to satisfy at the same time:

* How does a extension running the old version of GMailJS get its eventhandlers called (since it doesn't know about t he XHR singleton and haven't registered for callbacks), when a new version puts the Singleton implementation in place?
* We *could* prepare for an incremental migration, by having the patched (GmailJS-instance) XHR function have a property to let us know which instance it belongs too. This means that when we actually extract the XHR patching into its own singleton... We can know who we should automatically register.
* This approach depends entirely on one thing: All Gmail-extensions using GMailJS are being actively maintained. I honestly doubt that.
* Which leaves us the final question: Would it be acceptable to break extensions (but not Gmail!) running older GmailJS versions, with the fix being telling people to bugger their extension-author/maintainer to upgrade GmaiJS?

*phew*

That's a whole lot of concerns, en equal amount of work, quite a bit of complexity (and thus risk) and not to mention it would take quite a bit more testing than we usually do (with regard to cross-version compatibility testing)! This would also have to be a sustained effort over time (due to needing a gradual migration).

If I were to be completely honest, I'm not sure I think the value-proposition in removing the Singleton is worth all this. I'm certainly not going to bother being the person driving, designing, implementing and testing this. Maybe simply researching why chain-patching fails would be a better spent effort?

If anyone disagrees, or thinks I've overlooked a simpler way (and thus are over-complicating matters), feel free to chime in.

In the absence of any such comments, I'm inclined to close this issue and PR though.

patrickriordan:
@josteink I'm ok with closing this. Thanks for considering it! Devs should remember that if they run in a browser with another extension running an old version of gmailjs their api's could be broken.

josteink:
Thanks for the feedback @patrickriordan!

I was considering closing this myself, but it always feel better when the original poster does so himself.

In case we revisit this topic in the future, I just want to make absolutely clear that I don't *mind* this kind of change (In fact, while not having time to do it myself, I strongly encourage it!).

My only gotcha is that I'd like to see all concerns about having a cross-version compatibility  migration strategy gets addressed in a reasonable way. I'm sure it can be done, but like I said it will probably take a little bit of sustained effort.

josteink:
@patrickriordan: A similar change has now been comited in master via the following commits:

* https://github.com/KartikTalwar/gmail.js/commit/537038bf03f80c4954f494835a63f6a41ce628bd
* https://github.com/KartikTalwar/gmail.js/commit/6a11c174e15b47805b10e26f9e1070c78fcd3191

GmailJS is no longer a singleton, it should be safe to run as a non-singleton, because it should support nested XHR-patching, and it should support various concurrent GmailJS versions loaded via different means at the same time... In a backward-compatible manner.

All in all, I think that is a massive change which should make everyone happier.

Issue #362, opened 2016-12-28
current_page() function get wrong result 


HangYang:
Go to another page in some label like page 2 in sent label, the URL will be
https://mail.google.com/mail/u/0/#sent/p2

The[ first line in the function](https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js#L1758) will get "sent/p2", and the function will return
"email", which should return "sent".

fix:
change the first line from 
`var hash = window.location.hash.split("#").pop().split("?").shift() || "inbox";`
to 
`var hash = window.location.hash.split("#").pop().split("?").shift().split("/").shift() || "inbox";`

HangYang:
Did

josteink:
Fixed in https://github.com/KartikTalwar/gmail.js/commit/a89afcb68b9477efd1dca2b8b83d22aae5ef7236.

PR #363, opened 2016-12-30
fixed the bug in api.get.current_page

HangYang:
fixed issue #362

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1755,7 +1755,7 @@ var Gmail_ = function(localJQuery) {
     };
 
     api.get.current_page = function() {
-        var hash  = window.location.hash.split("#").pop().split("?").shift() || "inbox";
+        var hash  = window.location.hash.split("#").pop().split("?").shift().split("/").shift() || "inbox";
         var pages = [
             "sent", "inbox", "starred", "drafts", "imp", "chats", "all", "spam", "trash",
             "settings", "label", "category", "circle", "search"


josteink:
Thanks for the PR!

Initially looks good, but could you also please make 'window.location.hash' a intermediate variable which we could also pass in as an argument to support unit-testing?

Then you could add some simple tests like this and we'd be protected against future regression:

https://github.com/KartikTalwar/gmail.js/blob/master/test/test.parsing.js

Sounds doable?

josteink:
Sorry about breaking your fix, but I saw that making things unit-testable was a task in itself, so I did that work for you.

If you can resolve the conflict, we can work on getting this merged :)

josteink:
Looking into this, I found that the current implementation was too messy and that your patch alone wasn't enough to fix it for all use-cases.

I refactored the code while adding some new test-cases, on top of your changes. You can see those changes here:  https://github.com/KartikTalwar/gmail.js/commit/a89afcb68b9477efd1dca2b8b83d22aae5ef7236

From what I can tell, this function now works as it should. Feel free to let me know if you find some test-cases which

1. aren't covered and/or
2. aren't working.

Anyhow: Thanks for contribution!

HangYang:
It's great, thank you.

josteink:
As I suspected, this was too simple to be the full story :)

Specific parts of `api.get.visible_emails` depends on `api.get.current_page` to return data formatted differently than what we do now:

https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js#L1639-L1655

This means I'll have to to do some more changes to preserve the page-number in the results to ensure correctness with the rest of the code.

That means `gmail.get.current_page()` *will* return `"sent/p2"` for page 2 and that kind of stuff. So when I get this properly fixed, and push a new release, this is the behaviour you should be targeting.

Cheers.

josteink:
This should conclude this issue:

https://github.com/KartikTalwar/gmail.js/commit/ed0f72bffa5ca63f4248548f4b0a960e05218204

PR #366, opened 2017-01-13
clarify meaning of the `view_email` observer

apopiak:
the `view_email` event is only triggered once for a conversation thread and (possibly) includes several emails.
The name is actually misleading and should probably be changed.


josteink:
Thanks for the PR!

While changing the name would be a breaking change (and thus a no go), I agree a clarification on how this event works is a general improvement. 

Consider this merged!

Issue #353, opened 2016-12-12
Using api.get.email_source() fails in MS Edge


josteink:
No really. If you want to write WebExtensions which supports all modern browers, MS Edge is a target too.  :)

Basically something fails because of XHR with bad HTTP methods, and while it may not be fatal, the way we handle it causes it to fail:

* `api.get.email_source()` requests something from https://mail.google.com.
* This results in a 302 redirect.
* The new URL gets probed by jQuery using a HTTP OPTION request.
* This is where it fails. If we could *catch* this error, and just retry with a GET, we should be cool.

You can see this in the developer console here:

![image](https://cloud.githubusercontent.com/assets/411338/21090384/b611bd1a-c03f-11e6-8806-39e3b7f98a95.png)

![image](https://cloud.githubusercontent.com/assets/411338/21090416/e83bba98-c03f-11e6-863d-c682a2c06ac7.png)

@michi88: Anything you think you could contribute on this issue? You've touched this very specific code yourself :)

**Tip:** If you need help enabling extensions in Edge, [check the uBlock Edge docs for how to enable developer mode](https://github.com/nikrolls/uBlock-Edge#installation).

josteink:
This might actually be an issue in Edge itself:

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/9869121/

michi88:
We did have some CORS issues briefly discussed here: https://github.com/KartikTalwar/gmail.js/pull/333

This is related to the previous redirect being cached by the browser. This request needs to hit google as they will not add CORS headers to the download location otherwise. They open some kind of session or something.

After I've added jQuery's `{cache: false}` which adds something like this to the url `_=76348726487` it started working.

Maybe you can try to disable cache in Edge and see if that make a difference?

josteink:
Status update: From bad to worse.

First: I've now encountered this same behaviour in Firefox too.

Second: According to a network-specialist at my company, this behaviour (issuing a preflight OPTIONS request to check CORS-settings when redirected to a new domain) is actually the *correct* thing to do, and the behaviour in Chrome (which works for us) should be considered a bug.

As long as Gmail keeps it's current behaviour (returning a 405 when getting the OPTIONS request), this means we might soon see this behaviour in all browsers, and thus `email_source()` in all its variants will stop working.

As such, I think it would be wise to investigate a work-around for Firefox and EDGE, because we might actually *need* it to having a working library for Chrome in the not to distant future.

[...]

PR #377, opened 2017-02-20
Fix CORS issues for api.get.email_source()

josteink:
This commit does quite a few things all in one, which is counter to regular "git philosophy", but all in all, I think they belong together.

The changes can be summed up as:

1. Deprecate `api.get.email_source()` - Displays warning for now. Will remove it once we hit 0.7.0.
2. Added new function for downloading "troublesome" content like attachments and email MIME-source. This function is currently *only* used for this.
3. Add extra parameter to `api.get.email_source_async()` to allow downloading in binary form to avoid introducing encoding errors in MIME-messages.
4. Started migrating to more ES6ish constructs. Arrow-functions, `const`s, block-scoped `let`s etc. Updated corresponding linting/compilation settings. It was about time :)
5. Updated documentation.
6. Updated typescript-definitions.
7. Bumped version.
8. Updated changelog.

*phew*

Downloading email source is now verified to work in ALL modern browsers (Sorry Safari!). This fixes https://github.com/KartikTalwar/gmail.js/issues/353.

@michi88 , @KartikTalwar : I'll await your response before doing anything else (merging this to `master`) Ideally, this being a PR and all, I'm not the one doing the merging :)

Noteworthy: The part about introducing ES6 concepts will make the library *slightly* inconsistent API-wise when this is merged. Therefore, once all other issues have been settled, and this PR has been merged, I'd like to brush up some other parts of the library to also use more modern JS-concepts:

* Also provide Promises everywhere we have currently async-methods.
* Make consistent usage of `const` and `let` where appropriate.

Sounds OK?

(I'd also *love* to separate out the singleton implementation thing into its "own thing", so that we can convert Gmail.js itself to use proper classes, and what not. But I guess that will be its own big project some time in the future, *if* I have time and incentive to do so.)

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -4,6 +4,8 @@
 // https://github.com/KartikTalwar/gmail.js
 //
 
+/*eslint-env es6*/
+
 var Gmail_ = function(localJQuery) {
 
     /*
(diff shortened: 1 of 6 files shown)


KartikTalwar:
👍 for es6

src/gmail.js
```diff
@@ -2019,12 +2082,23 @@ var Gmail_ = function(localJQuery) {
...
+        const url = api.helper.get.email_source_pre(email_id);
         if (url !== null) {
-            api.tools.make_request_async(url, "GET", callback, error_callback, true);
+            return api.tools.make_request_download_promise(url, preferBinary);
         } else {
-            callback("");
+            return new Promise((resolve, reject) => {
+                // I honestly don't think this makes sense.  We really
+                // should reject the promise, but that would break
+                // backward compatibility W.R.T. _async...
```
KartikTalwar: I agree. In the next version bump we should reject and aim for consistency

KartikTalwar: the rest of the pr lgtm

josteink: If I were to be retro-critical to my own commit, and you agree we should aim for the more proper reject-semantics in an upcoming release anyway...

I think we could start aligning towards that goal right now by doing the right thing(tm) in the new `_promise` APIs (while leaving `_async` instact, for compatibility). How about it?

KartikTalwar: Let's do the right thing™

josteink:
Two updates in these last commits:

**First:** Considering how we *are* deprecating and telling people to migrate from `api.get.email_source` to `api.get.email_source_async` or `api.get.email_source_promise`....

I decided it would be better to just introduce the right semantics right away, rather than having duplicate code to maintain the "old" behaviour just for `api.get.email_source_async`. If we kept the old code there, and tried to fix it up later, then people migrating from synchronous code would effectively have to migrate twice once we decide to fix `_async` to behave as `_promise` later.

And that would just be silly. So I went ahead and did that change sooner rather than later. An error is an error. No shame in blowing up :)

**Second:** Also, seeing as the `api.get.email_source_*()` functions now accept a boolean argument after the emailID (and callbacks), it's very easy for callers to just pass in `null` for `emailID` expecting that to result in automatic resolving. The existing code didn't really do that (only when `emailID` was `undefined`), so I decided to fix up that as well.

Hopefully everything is still OK. If so, go ahead and merge at will :)

KartikTalwar:
This is great!

josteink:
So I guess this one is good to merge, then?

KartikTalwar:
@josteink yes, other than this travis warning (which I think is just a one off thing), everything is good!

josteink:
And done. Awesome!

I've been super-eager to get this one done for several weeks now. Good to finally be able to close it :)

michi88:
Awesome @josteink, great work!

Issue #247, opened 2016-04-08
visible_emails thread limitations


joydesigner:
Hi, 

gmail.get.visible_emails returns a maximum of 45 threads. Usually, gmail inbox view contains 50 emails per page.

Has anyone encountered similar issue? is there any workaround for this?

joydesigner:
Fixed. the url param for the request for inbox currently is "mbox", for enterprise gmail inbox, it should be "inbox".

josteink:
Sorry about the *extremely* slow response.

From what I can tell, you had an issue, and found a fix. Would you please consider submitting this fix as a PR? I promise I'll *personally* review it and I promise it won't take a year this time :)

Anyway, since there haven't been any activity since the time of filing, I'll assume this is no longer and issue and close it. If you still need something done, see you in that PR, ok? :)

PR #387, opened 2017-04-06
replacing query param "mbox" (for enterprise gmail account, apparently) with "inbox"

Banjerr:
I ran into an issue described [here](https://github.com/KartikTalwar/gmail.js/issues/247), where `visible_emails `would only pull back 45 results; switching `mbox` with `inbox` fixed the issue.


KartikTalwar:
Hey @Banjerr thanks for the PR! The comment in #247 says:

> for enterprise gmail inbox, it should be `"inbox"`.

Is the query param only a problem for google apps accounts or is it also a problem for gmail.com accounts?

Banjerr:
Hi @KartikTalwar 

Good question. I was testing this on my work account, which I believe is a google apps account, but just tested on my personal account (which I *believe* is a gmail account) and it had the same behavior. So, I'm not entirely sure, but I believe it's an issue for both

KartikTalwar:
I tested it locally and seems to work for me. In case anyone else has issues, please bring it up as a new issue - merging for now

josteink:
Good approach. No need to leave PRs hanging needlessly :)

PR #393, opened 2017-04-26
Regex - hyphen must be last or escaped by a backslash

JerryBels:
This bugs the function

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -2212,7 +2212,7 @@ var Gmail_ = function(localJQuery) {
 
 
     api.tools.extract_name = function(str) {
-        var regex = /[a-z"._-\s]+/gi;
+        var regex = /[a-z"._\s-]+/gi;
         var matches = (str) ? str.match(regex) : undefined;
 
         return (matches && matches[0]) ? matches[0].trim() : undefined;


josteink:
Can you come up with a concrete example of such a bug?

To me it looks like it may be possible to add a unit-test for that scenario. Could you add that?

JerryBels:
Hey @josteink , tbh I had these bugs two months ago when I worked last on my app, didn't have time for it since... and can't remember what made the bugs occur, only TODO'ed myself "push it to gmail.js"...

If I remember what it was about I will come back to say it. But anyway, a hyphen in a middle of a regexp is an error (even if javascript seems to tolerate it more often than not), and should be escaped with a backslash (actually it's even better than what I did by putting it last, some Microsoft tools rejects hyphen last and not escaped with a backslash).

josteink:
Hey Jerry.

Thanks for the feedback! Taking a closer look at this code, it actually seems to have more bugs/lackings, and the one you've found (IMO) only being a minor one.

The biggest issue I see with this function in its current form is that it won't handle non-ASCII characeters in a name (and they DO exist. I, Jostein Kjønigsen, insist so).

For now, it seems the impact of this bug is pretty small, and its only usage is inside ` get_displayed_email_data_for_single_email`, which again is only used by `api.get.displayed_email_data`.

I vote we try to fix this function, but if we do, we should do it properly, and have real-world test-cases in the test-suite.

JerryBels:
One more vote here ! But well... I don't really have time to do it right now ^^

josteink:
Trying to see how this errors out in the real world, I've tried invoking `api.get.displayed_email_data()`, but every single time, no matter what I do, the check on `api.check.is_conversation_view()` returns `true`, making the codepath which uses this code completely inaccessible.

So from what I can see, users may also need to disable a global gmail default-preference in order to trigger this code.

This bug just went from "low" to "lower" :)

josteink:
Ah. There we go:

image

Buggy indeed.

And then input we get does look strange too:

image

Anyway: This should be enough for me to construct a test-case, and ensure we pass properly (even though I may not do it right away).

I probably won't merge this PR as is, but I'll keep it open, since we're already using it to track a relevant bug :)

JerryBels:
Ah, you're fast ^^ Feel free to close the PR whenever you got the proper fix ready !

josteink:
Adding to the conversation, this was added by @g8g3 with this PR before he deleted his account:

https://github.com/KartikTalwar/gmail.js/pull/51/files

josteink:
Fixed in https://github.com/KartikTalwar/gmail.js/commit/886f3e7cf34cb88de010d8bd3153ef21315872d2.

Thanks for the PR. See you around :)

PR #397, opened 2017-05-08
Adds a feature to create a button on the right hand side of the toolbar.

srastanka:
Adds a feature to create a button on the right hand side of the toolbar at the end, right after the settings button.
On the other hand, the add_toolbar_button function adds a button only in the left half of the toolbar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,7 @@ These methods return the DOM data itself
 - gmail.dom**.get_left_sidebar_links()**
 - gmail.dom**.search_bar()**
 - gmail.dom**.toolbar()**
+- gmail.dom**.right_toolbar()**
 - gmail.dom**.compose()** - compose dom object - receives the DOM element for the compose window and provides methods to interact
 - gmail.dom**.composes()** - retrives an array of `gmail.dom.compose` objects representing any open compose windows
 - [gmail.dom**.email()**](#gmaildomemailemail_el-or-email_id) - email dom object - receives an email DOM element or email id for an email currently being viewed. Abstracts interaction with that email.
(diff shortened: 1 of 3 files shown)


josteink:
Hey @srastanka.

Thanks for the pull!

I'll be honest and admit I haven't tested these changes yet in a real gmail-project, but initially scanning the code I see you're following patterns and conventions from the existing code-base. And that's usually a good thing.

In my opinion however, I think there seems to be quite a bit of code-duplication.

From what I can tell the `toolbar()` and `right_toolbar()` shares almost *everything* except for two things:

* the DOM selector
* the default button-class if no style-class is provided

We can extract what's common into a internal "generic" toolbar function, and just dispatch to that one from the actual API methods. I've outline this in the quasi-code below.

````javascript
var create_generic_toolbar_button = function(params, cssStyle, selector) {
    // all the common stuff
};

api.tools.add_toolbar_button = function(params) {
    create_generic_toolbar_button(params, left_style, left_selector);
}
api.tools.add_right_toolbar_button = function(params) {
    create_generic_toolbar_button(params, right_style, right_selector);
}
````

Doing so would eliminate most of the code-duplication, and I think that's a worthy goal. If you could do that, I'll try to take a deeper look and do some actual testing once landed.

Sounds good?

srastanka:
Hi @josteink, 
Thanks for checking out my PR and for your comments.

You brought up a very good point about the code duplication. I agree with your comments.
But would like to mention that there are actually three differences:

- the DOM selector

- the default button-class if no style-class is provided

- the base class is also different for the two styles (`"T-I J-J5-Ji lS"` for the left and `"T-I J-J5-Ji ash"` for the right)

So I've refactored the code to follow proper coding practice and put the common code in the `create_generic_toolbar_button` function and then dispatching to that from the actual api methods.

Let me know how does everything look and if you think anything needs modification.

Thanks.

srastanka:
Hi @josteink .

Have you had a chance to look at this and do some testing?

josteink:
From an initial glance it looks good. 

I'll try to get some testing done tomorrow.

josteink:
Seems to do what it says, and I don't see any obvious problems with how it works either!

I forgot to request one final change though, but if you can do that as well, I'll merge this in an instant without any further delay.

Basically when adding a new function we need to also do *two* things:

* Add it to the documentation (which you did. Good good!)
* Add it to the TypeScript type-definition. This is still lacking.

You should pretty much be able to duplicate the following line, rename the function-name in the copy and you should be home free: https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.d.ts#L650

If you can do that, I'll get this merged once and for all. Thanks for your contribution, and again sorry for the hold-up!

srastanka:
Hi @josteink .
Sorry for the delay. I've added the function to the TypeScript type-definition.
Please check and merge.

josteink:
> if you can do that as well, I'll merge this in an instant without any further delay.

Looks good to me, and as such I'll honour my word.

Thanks for the patch and thanks for contributing to `gmail.js`!

PR #426, opened 2017-09-24
Chrome extension

rushglen:
I've removed this because my code was in further testing not working 100%
Sorry for the mess I've created!


josteink:
There seems to be some errors with this PR. The diff doesn't make any sense, and it doesn't contain any actual commits made by you.

Can you try recreating one from scratch, and maybe use Github's web-editor to make your changes if you're unfamilier with git's command-line interface?

KartikTalwar:
It also appears that on top of updating the core library functions, the PR is adding the sample chrome extension to this repo as well. I'd suggest that we keep the chrome extension code to the [boilerplate repo](https://github.com/KartikTalwar/gmail-chrome-extension-boilerplate)

PR #427, opened 2017-10-01
Add CDNJS and npm version badges in README.md

sufuf3:
These badges will show the version on CDNJS and npm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Gmail.js - JavaScript API for Gmail
 
 ![Build status](https://api.travis-ci.org/KartikTalwar/gmail.js.svg?branch=master)
+[![npm](https://img.shields.io/npm/v/gmail-js.svg)](https://www.npmjs.com/package/gmail-js)
 
 ### What Gmail.js is and isn't
 


josteink:
Hi there and thanks for the patch!

While I’m fine with having badges in general, I have 2 questions:

* is shields.io an officially sanctioned way to get a NPMJS badge? (Similarly for CDNJS)
* should we encourage use of CDNs for libraries used inside browser extensions? I think that sounds very strange and I think we may be better off without it. 

Any opinions, @KartikTalwar ?

KartikTalwar:
Having the latest version on npm definitely makes sense but I agree, it might be confusing for others to see gmail.js on cdnjs and wonder how to use it since it only works for chrome extensions. I don't have any strong objections to it, but do believe that it'll lead to confusion

sufuf3:
@josteink @KartikTalwar Thank you for your suggestion.
IMO, the front-end developers know how to use web public CDN and add this badge will help them know more information.
So I think maybe it's good to have.

But if you think it's not necessary I'll update this PR, let it better to merge.
What do you think?

Thank you very much.

josteink:
I guess that settles it: NPM badge is fine. CDNJS might confuse people.

sufuf3:
Removed the CDNJS badge part
Please help me review this PR, Thanks. :)

josteink:
Unless @KartikTalwar has any objections, I'll be absolutely happy to merge this.

It looks nice :)

KartikTalwar:
👍

josteink:
Thanks for the patch. Merged!

Feel free to submit more in the future!

sufuf3:
Thank you! :)

PR #432, opened 2017-10-20
Try to improve locale-detection.

### LINKED ISSUE #430: Gmail.js does not always use correct globals[17] sub-array for locale-data
Currently our code checks for values in globals[17][8] or globals[17][9], but I've found at least one case in the wild with globals[17][7].

We need our code to be more robust against such changes.

### LINKED ISSUE #431: Gmail.js extracts a value used by google to denote user-origin, conflates with UI language
As seen in quite a few cases, Gmail now uses `globals[17][x][8]` to denote the user origin, and not the UI language,

This means that a norwegian user, with a english UI Gmail will receive a norwegian locale.

This is not expected behaviour.

josteink:
As noted in the code, the globals[17] array is not entirely reliable with regard to index-based access.

I've therefore created some helper-functions to asses which sub-array is the correct one based on its key-value (which consistently has shown up as `"ui"`).

I've also applied a more intelligent approach by looking for other Google links with explicit language-codes for UI embedded in them, rather than assuming that the user's denoted locale is always applicable as a UI-language.

These language-codes *can* contain region-denominators (en-GB, etc) and for the sake of backwards compatibility, these gets filtered out.

I'm pretty sure this is 100% clean, but review would still be nice before pushing to master.

**Edit:** I've added more commits. This commit fixes #430 AND #431.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -142,9 +142,22 @@ var Gmail_ = function(localJQuery) {
             return false;
         }
 
+        if (locale.match(/[0-9]/)) {
+            return false;
+        }
+
         var localePrefix = locale.slice(0, 2);
         return localePrefix.toLowerCase() === localePrefix ||
             localePrefix.toUpperCase() === localePrefix;
+    };
+
+    api.helper.filter_locale = function(locale) {
+        if (!api.helper.get.is_locale(locale)) {
+            return null;
+        }
+
+        // strip region-denominator
+        return locale.substring(0,2).toLowerCase();
     };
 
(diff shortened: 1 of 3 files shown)


josteink:
cc: @KartikTalwar

josteink:
I need this in a production project, so if I don't receive any feedback in a few more days, I'll just go ahead and merge.

IMO the changes proposed here are perfectly safe and with the included resiliency checks and tests, I would consider the risk absolutely minimal.

josteink:
No comments and 10 days passed. I'm merging this.

Issue #379, opened 2017-02-24
We should review our jQuery-dependency


josteink:
Currently we say we depend on jQuery 1.10. This is also declared in our [package.json](https://github.com/KartikTalwar/gmail.js/blob/master/package.json#L25-L28)-file, meaning that is what a `npm install gmail-js` will pull in.

But this version is *ancient*. It's from [November 15th, 2013](https://jquery.org/history/) and [supports MSIE6](https://blog.jquery.com/2014/04/02/browser-support-in-jquery-1-12-and-beyond/) for crying out loud!

Are there any technical reasons why we should discourage using a modern version of jQuery with this library? The only browsers we need to support are up to date Firefox, Chrome and MS EDGE. All of which are all supported by [latest jQuery](https://jquery.com/browser-support/).

As far as I can see, upgrading shouldn't represent any obvious issues, but I'd like to double-check before doing any "radical" changes with regard to core dependencies :)

Anyone for? Anyone against?

KartikTalwar:
No objections. Since selectors and ajax are the only thing we rely on, I don't think there should/would be any problems with using the latest jQuery

josteink:
The only thing I could think of, if I were to try to think *real hard* is that we use jQuery for AJAX requests. In *theory*, there could be some change in behaviour which may or may not work out well for us.

I think if we test that a few basic operations (`api.get.email_data()` etc) works out, it should be fairly risk free.

I'll try to get this batched in and released together with the CORS fix before pushing a new version to npm.

josteink:
Doing some basic testing, I get this:

![image](https://cloud.githubusercontent.com/assets/411338/23316404/8c94d0c4-faca-11e6-8f7d-18e1c2c0ba51.png)

Basically it works, but something new inside jQuery decides to try to interpret everything as JSON, no matter what. Failing at that, it makes ugly log-statements in the console.

![image](https://cloud.githubusercontent.com/assets/411338/23316446/c0b0c764-faca-11e6-93d8-6067ede630ba.png)

If I were to take a guess, it would probably cause less panic among library-users, if we figured how to avoid triggering those errors before making the change. I'll defer it to later, I guess.

josteink:
So almost one year passed, and now GitHub is flagging the version of jQuery we ship as insecure:

![image](https://user-images.githubusercontent.com/411338/35260948-458ac95a-000e-11e8-9f96-b15a3f85d3fc.png)

[...]

PR #436, opened 2018-01-31
Fix ajax request error.

Navega:
Fix ajax request: "SyntaxError: Unexpected token )", we need to specify the datatype for the response.

I am receiving a lot of errors in the console due to that problem. It isn't critical, but I recommend to fix it.

Navega, Software Developer at Appfluence Inc.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1637,7 +1637,7 @@ var Gmail_ = function(localJQuery) {
         method  = method || "GET";
 
         link = encodeURI(link).replace(/#-#-#/gi, "%23");
-        var config = {type: method, url: link, async: false, dataType:'text'};
+        var config = {type: method, url: link, async: false, dataType:"text"};
         if (disable_cache) {
             config.cache = false;
         }
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1637,7 +1637,7 @@ var Gmail_ = function(localJQuery) {
         method  = method || "GET";
 
         link = encodeURI(link).replace(/#-#-#/gi, "%23");
-        var config = {type: method, url: link, async: false};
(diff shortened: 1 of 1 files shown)


josteink:
Hey there, and thanks for another patch :)

My initial response to this is that this looks fully reasonable. Code-wise, there's some style-issues you'll need to fix (double-quotes, not single), but as far as I can tell, that's about it.

It we're lucky, this may also fix the problems we have with https://github.com/KartikTalwar/gmail.js/issues/379.

Before merging though, we will need to quickly map out how much calling code is impacted, and what kind of testing we will use to ensure that we're not breaking anything.

Navega:
Hey there,

I fixed the code style, thanks!

I just saw that you use the same "dataType" config in the function "make_request_async", so I am very confident with the fix, and our QA department tested it our Gmail Extension and it works fine with the new fix. It isn't critical, the only real problem is a parsing jquery action which is sending error messages to our loggers.

If you need something from me, only tell me.

josteink:
Sounds good. I don't think I'll need much more help on your part.

That said: I'd like to do some basic end-to-end testing on a complex extension I'm responsible for before committing, and I'm not sure when I'll have time to do that. Quite busy at work now...

If you feel things are taking too long time, don't be afraid to remind me :)

Navega:
Ok, thanks @josteink!!

josteink:
I've never had this issue with the (ancient) version of jQuery we've shipped by default, but when testing with a modern version (as in #379), I can reproduce the issue.

Testing my most sophisticated and intricate extension using your patch and latest jQuery, I get no warnings and everything works fine.

Thanks for the patch. Feel free to keep them coming :)

PR #437, opened 2018-01-31
Fix: parameter "sstart" for offset.

Navega:
Fix: now the parameter "sstart" is Nan, so the offset is broken, so we need to split it by "–". (i.e: "51–100").

image

Navega, Software Developer at Appfluence Inc.


josteink:
I can reproduce the problem and verify the fix.

Thanks for another good patch!

Issue #429, opened 2017-10-15
insertion_observer 'cn.trim is not a function'


revmischa:
When hiding gmail native compose:
<img width="734" alt="screen shot 2017-10-15 at 14 32 12" src="https://user-images.githubusercontent.com/245131/31584793-ae9175f4-b1b5-11e7-9eac-5262b0e53efe.png">

josteink:
Could you elaborate a little on this error. Can you provide some sample code to reproduce it?

revmischa:
I added some debugging and found that it is erroring on this:
<img width="682" alt="screen shot 2018-02-15 at 11 23 07" src="https://user-images.githubusercontent.com/245131/36252787-7178a38c-124e-11e8-80e5-62e670cc7e65.png">

josteink:
Old issue is old. Closing.

PR #443, opened 2018-02-15
#429 some classNames maybe SVGAnimatedStrings and not have trim()

revmischa:
#429

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1582,7 +1582,7 @@ var Gmail_ = function(localJQuery) {
 
         // loop through each of the inserted elements classes & check for a defined observer on that class
         var cn = target.className || "";
-        var classes = cn.trim().split(/\s+/);
+        var classes = cn.trim ? cn.trim().split(/\s+/) : [];
         if(!classes.length) classes.push(""); // if no class, then check for anything observing nodes with no class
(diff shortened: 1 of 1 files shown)


josteink:

I got a few questions though, before we merge this. Hope it's ok.

1. Could you outline a step-by-step procedure to reproduce the error in #429? While there's lots of screenshots showing that something *has* gone wrong, it doesn't completely explain *how* to get there. This would definitely help to verify and test the patch.
2. This patch avoids the error condition by detecting it up front, and inserting empty data. What does that do to the rest of the code? Does it cause side-effects down the line?

Should we instead abort here, or is this indeed correct behaviour?

cc: @KartikTalwar

revmischa:
We're getting notified of things with animated classes being inserted. Hopefully no gmail event observers depend on those.

I don't know how to reproduce this yet with a minimal test case. If I find out I'll let you know.

josteink:
I can't reproduce this, thus can't directly verify the fix. The change itself *does* look fairly harmless though.

Without this fix, the code would have crashed either way... So in worst case... If this fix causes another error down the line... Technically speaking we shouldn't have broken anything which wasn't already broken.

Since I can't see anyone else objecting (nor reviewing really), I'll merge this for now.

Thanks for the patch!

Issue #381, opened 2017-03-08
False positive on delegated inbox detection


remidenele:
Hi, I have detected a problem with the [`gmail.get.delegated_to_email()`](https://github.com/KartikTalwar/gmail.js#gmailgetdelegated_to_email) function which is defined as follow:

> Returns the email address of the user the account is currently delegated to (if the inbox is used by the owner, this function returns null)

It is working as described in the doc on one of my G Suite domain where delegation is enabled, but not on an other one where email delegation is disabled at a domain level:

![mail-delegation-management](https://cloud.githubusercontent.com/assets/8693949/23706739/b4b8d494-040f-11e7-925a-a1887bb98d76.png)

On the 2nd domain (mail delegation off), [`gmail.get.delegated_to_email()`](https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js#L113) always returns the mailbox owner email address instead of `null`.
It appears that [`api.helper.get.is_delegated_inbox`](https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js#L1683) is always returning true in that case.

josteink:
Just to be sure... Have you tried completely logging out and reloading your browser, Gmail (with your extension) and all those things after changing the domain settings? Sometimes, some of these things can be sticky.

If reloading fixes it, I'm not going to consider this a big deal.

If it's a persistent issue however, it's definitely something where we should take a look into possible means of action.

remidenele:
In fact, I have not changed domain settings: I am using 2 different domains with different settings:
- _domain**1**.com_ where Mail Delegation setting is **enabled** at the domain level
- _domain**2**.com_ where Mail Delegation setting is **disabled** at the domain level

I am using 2 distinct Chrome profiles. On each profile, only one G Suite account is logged in in Gmail (no multi login used in Gmail), and none of the mailboxes are shared (no delegate):

- for _user-test@domain**1**.com_, `gmail.get.delegated_to_email()` returns `null` (behaves as described)
- for _user-test@domain**2**.com_, `gmail.get.delegated_to_email()` returns  `user-test@domain2.com`

For now, my workaround is to add a check on the value returned by `gmail.get.delegated_to_email()` to ensure it is not equals to the current logged user.

[...]

PR #446, opened 2018-02-26
#381 - Fix possible false positive delegate inbox detection

remidenele:


josteink:
This looks reasonable to me.

Could you update the [CHANGELOG](https://github.com/KartikTalwar/gmail.js/blob/master/CHANGELOG.md) to contain a description of your fix?

josteink:

This change alone looks absolutely reasonable to me, and I'd be happy to merge it.

But before we do that, I'd like you to do two things:

* Could you update the [CHANGELOG](https://github.com/KartikTalwar/gmail.js/blob/master/CHANGELOG.md) to contain a description of your fix? Add it under a new version (0.6.14).
* Could you amend this PR with your other change (and then close it), so that they can be reviewed and tested together?

remidenele:
I close this one as it should no longer be necessary if we fix api.helper.get.is_delegated_inbox as described in #447

Issue #445, opened 2018-02-26
api.helper.get.is_delegated_inbox always returning false


remidenele:
The [below code](https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js#L1761) used to detect if the current mailbox is used by a delegated account is no longer working:

```javascript
api.helper.get.is_delegated_inbox = function() {
    return api.tracker.globals[17][5][0] === "fwd";
};
```

It seems that `api.tracker.globals[17][5][0]` is equals to `acfg` in case a delegated account is in use and to `dlg` (or `md:tes` on some accounts) when the manager account is used (or when it is a "simple" account with no delegation)

**Note**: `acfg` is also returned when delegation is not allowed at all at a domain level (cf. #381)

remidenele:
Fixed in 0.6.14. 
Thanks @josteink for the merge :+1:

PR #447, opened 2018-02-26
#445 - Fix api.helper.get.is_delegated_inbox check

remidenele:


josteink:
I'm running a normal Google Apps account with no delegation and when I test, I get the following:

````javascript
api.tracker.globals[17][5][0]
// "acfg"
````

The existing code may not work, but using your proposed code there is no way (at least in my case) to differ between *disabled* delegation and *active* delegation.

And if that's the best this function can do... That's not very useful really.

Could anyone else than me test and chime in?

@psobiech: As far as I can tell, you originally commited this code. Would you care to help out?

**Comment by josteink** [comment]
**Update:** I see you've submitted 2 PRs which contains invidual parts of the fix. That makes things a little bit confusing on my end :)

In the future please consider sending everything as one PR, so that it's easier to review.

**Comment by remidenele** [comment]
Indeed I submitted 2 PRs as the 2 problems look distinct to me, even if they deal with the same subject (inbox delegation).

I will regroup the changes after I will have performed more tests with other G Suite accounts as you seem to have a different result than mine for `api.tracker.globals[17][5][0]` value.

> I'm running a normal Google Apps account

Do you mean a G Suite account or a Gmail one? (I have only tested on 3 G Suite Business edition accounts)

**Comment by josteink** [comment]
> Do you mean a G Suite account or a Gmail one?

Sorry. GSuite. It used to be called "Google Apps" before Google renamed it for the gazillionth time. :)

**Comment by remidenele** [comment]
Here are my findings and tests:
[Gmail Inbox delegation tests](https://goo.gl/MxfU9d)

- `GLOBALS[17][5][0]` can be equals to one of the 3 values: `md:tes`, `acfg` or `dlg`.
- `GLOBALS[17]` array content varies depending on the context and more specifically on the accounts listed in the top right drop down menu
- if you have at least on delegated account listed in the accounts list (marked as "(delegated)"), `GLOBALS[17][5][0]` is equals to `dlg`
- if "Mail delegation" domain setting is disabled OR if you are logged in on behalf of someone else (delegated inbox), `GLOBALS[17][5][0]` is equals to `acfg`
- if you have no "(delegated)", and "Mail delegation" setting is not disabled on the domain, `GLOBALS[17][5][0]` is equals to `md:tes`

A bit complicated :-) and no sure we can still use `GLOBALS[17][5][0]` to achieve delegate detection.

I have notice a way to detect that the user in currently logged in as a delegate by parsing the DOM, but not sure it is reliable enough, especially knowing that css classes looks generated/obfuscated which is not the case for this one:

delegated-css

**Comment by josteink** [comment]
> Here are my findings and tests:

That's good testing and great work. Really appreciated.

So if this data is correct and representative for other GSuite accounts (will try to test myself) that means that:

* Regular Gmail and GSuite domain with delegation enabled looks the same (`md:tes`)
* GSuite domain with delegation enabled and disabled looks different. (`acfg` vs `ms:tes`)
* Active delegation can be detected. (`dlg`)

So if we can test for this being a gsuite account or regular gmail... We can accurately identify all three cases. Right?

> not sure we can still use GLOBALS[17][5][0] to achieve delegate detection.

This comment makes me wonder if there was a critical part I missed.

**Comment by remidenele** [comment]
> That's good testing and great work. Really appreciated.

You're welcome, I really appreciate your help and commitment to try to find a solution to this too!

> Active delegation can be detected. (dlg)

`dlg` just indicates that you have at least one delegated account in your accounts list but not that you are currently logged in a delegated inbox:
- column E, user-2 has access to a delegated inbox (user-3) but he is currently in his own inbox (user-2): `dlg`
- column F, user-2 is effectively logged in the delegated inbox (user-3): `acfg`
So we get the same value (`acfg`) either we are effectively in a delegated inbox or when delegation is disabled on the domain (column B).

So checking only the value of `GLOBALS[17][5][0]` to fix `api.helper.get.is_delegated_inbox` is not enough.

The only working solution I have so far is:

```javascript
api.helper.get.is_delegated_inbox = function() {
    return $('.identityUserDelegatedAccount').length === 1;
};
```

But I don't know if the DOM is stable enough to rely on this css class.

josteink:
> I don't know if the DOM is stable enough to rely on this css class.

I don't see how relying on DOM is any worse than magic constants in the `GLOBAL` variable. Besides, we're using DOM stuff lots of other places in the codebase.

If what we already have is 1. convoluted and 2. doesn't even work, and 3. we can find a reliable and obvious DOM-selector to work with instead... I'd rather have that as long as it works for now and does what is expected.

Do you have any objections, @KartikTalwar?

src/gmail.js
```diff
@@ -1759,7 +1759,7 @@ var Gmail_ = function(localJQuery) {
 
 
     api.helper.get.is_delegated_inbox = function() {
-        return api.tracker.globals[17][5][0] === "acfg";
+        return $(".identityUserDelegatedAccount").length === 1;
     };
```
josteink: Are we sure this will always be Length one? Are there cases you can see where we may have *more* than one?

Wouldn't it be better/more resilient to change to do a check if it's not zero?

josteink:
If that works, I have to admit I'm much happier with this approach. I'll be glad to merge that once we're done with review :)

src/gmail.js
remidenele: It is always 0 or 1 from my testings (on the different accounts / contexts listed in the spreadsheet), and it should always be as this selector identifies the logged in account in the accounts list when connected in a delegated inbox, so 1 max at a time.

I would say it is more a matter of "preference", either we prefer that this check fails if this css class starts appearing somewhere else in Gmail as it is not expected or we consider that if it appears more than once we are probably still in a delegated inbox context.

josteink: In that case, looks good to me.

If nobody else has any objections over the following days, I will merge this as is.

Thanks for the patch!

PR #471, opened 2018-05-10
Added support for "strict" TypeScript mode

SuperPaintman:
Greetings!

I added support for TypeScript's "strict" mode.

In this mode TypeScript requires a explicit definitions for return types and types of parameters.

### DIFF
--- a/src/gmail.d.ts
+++ b/src/gmail.d.ts
@@ -215,7 +215,7 @@ interface GmailGet {
     /**
        Does the same as visible_emails, but with a callback instead.
      */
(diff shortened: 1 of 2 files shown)


josteink:
Thanks! That’s a great PR. 

Upon my initial glance I can’t see anything wrong with it, but let me check it more closely on Monday when I will have more time. 

Cheers!

SuperPaintman:
Sure! Have a nice weekend!

josteink:
This was such a super-small patch, that I took the 5 minutes I needed to verify it right now.

Thanks for helping out with the TypeScript stuff! Consider this merged :)

PR #474, opened 2018-05-14
Fix 'before' trigger in new Gmail

phyxolog:


josteink:
LGTM.

Thanks for the patch!

PR #475, opened 2018-05-14
add archive and delete actions for new gmail dataLayer

hankypanky666:
example of usage
```
gmail.observe.after('archive', (oldIds, paramsUrl, paramsBody, newEmailIds, newThreadIds) => {
                if(!oldIds) {
                    postArchivedOrDeletedThreads(oldIds);
                } else {
                    postArchivedOrDeletedThreads(newEmailIds);
                }
              });
```
and we need mapping from new to old threadId


src/gmail.js
```diff
@@ -961,6 +961,26 @@ var Gmail = function(localJQuery) {
             if (pathname && (pathname.endsWith("/i/s") || pathname.endsWith("/i/fd"))) {
                 api.tools.parse_request_payload(params, triggered);
             }
+            if (pathname && (pathname.endsWith("/i/s"))) {
```
josteink: This whole following section should probably be put inside `api.tools.parse_request_payload()`, like other event's currently parsed.

josteink:
Hey there and thanks for the PR!

All contributions are welcome, and especially from new contributors!

That said, before we merge this, I think we will have to make some changes to the code in question.

Please see the comments I have on the code itself, and let me know what you think.

src/gmail.js
```diff
@@ -961,6 +961,26 @@ var Gmail = function(localJQuery) {
...
+                if (params.body_params && params.body_params[2] && params.body_params[2][1] && params.body_params[2][1][0][2][2][7]) {
+                    // check params length  and get thread ids
```
josteink: My experience trying to access specific indices on the request payload without further analysis has been less than impressive. For my use-case (sent email), it has often resulted in unreliable detection of events and meta-data.

For that reason I created the `api.tools.extract_from_graph()` function and associated graph-analysis-functions like `api.check.data.is_thread()` and `api.check.data.is_email()` (all of which are easily unit-tested).

Reading this code, it's not clear to me what kind structure we are looking for, and what kind of graph-entity it "is".

Would it be possible to rewrite this code to be more in line with how `api.tools.parse_request_payload()` already does things?

Together with that, if you have a screenshot of an example payload like this (to help illustrate what kind of request we're processing here) it will make it *much* easier to comment on the parsing code itself too.

src/gmail.js
```diff
@@ -961,6 +961,26 @@ var Gmail = function(localJQuery) {
...
+                    // console.log('params.body_params: ', params.body_params);
+                    if (params.body_params[2][1][0][2][2][7][2].length > 3) {
+                        const newThreadIds = params.body_params[2][1].map(obj => obj[2][1]);
+                        const newMessageIds = params.body_params[2][1].map(obj => obj[2][2][7][3]);
+                        // console.log(newMessageIds);
+                        // console.log('newThreadIds: ', newThreadIds);
```
josteink: These commented out logging statements should probably be removed once we have code in a finalized form which we see work well.

src/gmail.js
```diff
@@ -961,6 +961,26 @@ var Gmail = function(localJQuery) {
             if (pathname && (pathname.endsWith("/i/s") || pathname.endsWith("/i/fd"))) {
                 api.tools.parse_request_payload(params, triggered);
             }
+            if (pathname && (pathname.endsWith("/i/s"))) {
+                if (params.body_params && params.body_params[2] && params.body_params[2][1] && params.body_params[2][1][0][2][2][7]) {
+                    // check params length  and get thread ids
+                    // console.log('params.body_params: ', params.body_params);
+                    if (params.body_params[2][1][0][2][2][7][2].length > 3) {
+                        const newThreadIds = params.body_params[2][1].map(obj => obj[2][1]);
+                        const newMessageIds = params.body_params[2][1].map(obj => obj[2][2][7][3]);
```
josteink: What format are these IDs? Are they the new format as provided by new Gmail? Will these be compatible with the rest of our data-APIs which (for now) only supports old-style IDs?

If not... We need to make it clear somehow that the IDs provided with these events cannot be used with the rest of our APIs. Through documentation, or something.

If possible we should try to also locate IDs which are compatible with our old-style data-APIs, until they have been changed to support the new XHR data-formats.

src/gmail.js
```diff
@@ -961,6 +961,26 @@ var Gmail = function(localJQuery) {
...
+                        // console.log(newMessageIds);
+                        // console.log('newThreadIds: ', newThreadIds);
+                        if (params.body_params[2][1][0][2][2][7][1][0] === "^k") {
+                            // console.log("delete_new");
+                            triggered["delete_new"] = [newThreadIds, newMessageIds];
```
josteink: I don't think it makes sense to create events with *new names* just because it's a new version of Gmail underneath. We should just keep using `"delete"` and `"archive"` and the gmailjs library should abstract away this as much as possible.

For instance `"send_message"`-event has the same name for new and old data-layer.

src/gmail.js
```diff
@@ -961,6 +961,26 @@ var Gmail = function(localJQuery) {
...
+                        }
+                        if (params.body_params[2][1][0][2][2][7][1][0] === "^a") {
+                            // console.log("archive_new");
+                            triggered["archive_new"] = [newThreadIds, newMessageIds];
```
josteink: The old event provides the following event-argiments: `[email_ids, params.url, params.body]`

We should probably strive to have a similar looking argument-list to these events too.

That leaves us the problem that in the "old" Gmail email-ID and thread-ID could be used interchangably, but now they no longer can (which has benefits and drawbacks).

How about we make this argument list: `[email_ids, params.url, params.body, thread_ids]`

That should at least make old style event-handlers compatible with the additional event metadata.

Or does anyone foresee any problems with that?

src/gmail.js
```diff
@@ -1255,7 +1275,7 @@ var Gmail = function(localJQuery) {
                 }
 
                 // if any matching after events, bind onreadystatechange callback
-                if(api.observe.bound(events,"after")) {
+                if(api.observe.bound(events, "after")) {
```
josteink: This change is strictly speaking not needed, but I guess we can nitpick formatting-issues (if any) once everything else is in place 😄

src/gmail.js
hankypanky666: for this logic, we need mapping new threadIds with old threadIds

josteink: Unfortunately, I think the documentation given so far says there's no straightforward or reliable way to do that.

In my extension, the hacky way I've done this is simply to *navigate* to the new thread/email-ID and pick up the legacy-ID from the DOM. But that's hardly an approach we can use inside an event-handler.

That said: I still think it would be bad to create *new* events just because the data layer changes, and would rather not merge a PR with such a change.

Maybe we can make the new event-signature `[null, params.url, params.body, new_email_ids, new_thread_ids]`. That would have a backward compatible signature, with a way of checking for the presence of the legacy-id.

Iirc there has been a few different cases in GmailJS where the email-ID cannot be directly inferred and its up to the extension-author to determine what is the best approach for resolving that after the fact (or maybe by using a stored emailID in a closure, etc).

Leaving the legacy value `null` will give extension-authors the same flexibility here, at least until we upgrade our data-APIs or find another workaround for this problem.

hankypanky666: ok, this is a cool idea with a new signature

src/gmail.js
```diff
@@ -982,7 +982,7 @@ var Gmail = function(localJQuery) {
     api.check.data.is_email_id = function(id) {
         return id
             && typeof id === "string"
-            && /^msg-a:/.test(id);
+            && /^msg-[a|f]:/.test(id);
     };
```
josteink: These are good improvements and obviously something I missed first time around. Thanks for fixing!

src/gmail.js
```diff
@@ -1089,6 +1142,16 @@ var Gmail = function(localJQuery) {
                 }
             }
         }
+
+        if (threads[0]) {
+            actionType = api.tools.check_event_type(threads[0]);
+        }
+
+        if (actionType) {
+            const new_thread_ids = threads.map(thread => thread[1]);
```
josteink: This if-statement will only ever run if the previous if-statement has run, so it might as well be inside that if-block instead of in its own.

And if you do that, you can also make the `actionType` a block-level variable making the overall structure of the function cleaner.

src/gmail.js
```diff
@@ -1067,7 +1075,52 @@ var Gmail = function(localJQuery) {
...
+            ""            : "undo_send",
+            ""            : "unmute",
+            ""            : "unread",
+            ""            : "unstar",
+            ""            : "new_email",
+            ""            : "poll",
+            ""            : "refresh",
+            ""            : "restore_message_in_thread",
+            ""            : "open_email",
+            ""            : "toggle_threads"
```
josteink: I see what you're trying to do here, and it sort of makes sense. However if you *run* this code, it doesn't really work.

I suggest we comment out the unimplemented events for now, but keep them in as comments as a reminder of things which can/should be implemented later.

Does that sound OK?

src/gmail.js
```diff
@@ -1089,6 +1142,16 @@ var Gmail = function(localJQuery) {
...
             }
         }
+
+        if (threads[0]) {
+            actionType = api.tools.check_event_type(threads[0]);
+        }
+
+        if (actionType) {
+            const new_thread_ids = threads.map(thread => thread[1]);
+            const new_email_ids = threads.map(thread => thread[2][7][3]).reduce((a, b) => a.concat(b), []);
```
josteink: This code does feel a little "brittle" with regard to how the data is expected to be.

What happens if the structure changes? Will our XHR-intercept code crash and break Gmail? Could this be split into its own method where we are more careful to handle parsing errors?

josteink:
Thanks for having in there!

Looking over the recent changes, this is *much* better. I still have a few nit-picks, but I think we're getting close to something I can test and hopefully merge.

Let me know what you think :smile:

src/gmail.js
hankypanky666: yes)

hankypanky666:
well, I will continue to work tomorrow, thanks for the review!

src/gmail.js
```diff
@@ -992,6 +992,14 @@ var Gmail = function(localJQuery) {
             && api.check.data.is_email_id(obj["1"]);
     };
 
+    api.check.data.is_action = function(obj) {
+        return obj
+            && Array.isArray(obj)
+            && obj.length === 1
+            && typeof obj["0"] === 'string'
+            && !api.check.data.is_email_id(obj["0"]);
+    };
```
josteink: I’m still not sure what kind of structure this is. 

Having a unit-test for this function would both help ensure correctness *and* illustrate what kind of data we are working with.

hankypanky666: ok!)

josteink:
Thanks for the updates. I’ll try to take a look at them as soon as I can, which probably means around tuesday. Cheers!

josteink:
I see you've added lots of commits since last time.

Do you consider this ready for review now? Do you want me to take a look, or do you still have things you'd like to add?

src/gmail.js
```diff
@@ -1067,7 +1100,57 @@ var Gmail = function(localJQuery) {
...
+            const action = api.tools.get_action_type(threadData);
+            if (!action_map[action]) {
+                return;
+            } else {
+                return action_map[action];
+            }
+        }
+
+        return;
+    };
```
josteink: This part has been cleaned up considerably, and I really like that. This is much, much better.

Now though, I see it can be cleaned up even more 😄 

1. I think checking `if (action_map[action])` directly would be better, and avoiding the whole if-else business.
2. That means we only need one additional return, at the end of the function, as a fallback for other cases.
3. I'd also like that final `return`-statement to return `null` for consistency (either all code-paths return something or no-one does)

src/gmail.js
```diff
@@ -1089,6 +1172,23 @@ var Gmail = function(localJQuery) {
                 }
             }
         }
+
+        try {
+            if (Array.isArray(threads) && api.check.data.is_thread(threads[0])) {
+                actionType = api.tools.check_event_type(threads[0]);
```
josteink: `actionType` can be declared here, no need to have it declared up there.

test/test.tools.js
```diff
@@ -45,3 +47,40 @@ describe("Monkeypatching", () => {
         assert.equal("nested patched orig", result);
     });
 });
+
+describe("Test tools for parsing new gmail body_params", () => {
+    const gmail = new Gmail();
+    const data =  JSON.parse(testData.new_gmail_archive_action_body_params);
+    const mockMessageIds = ['msg-f:1600724307680265309', 'msg-f:1600724938213937205', 'msg-f:1600725174437456906', 'msg-f:1600725319255992336', 'msg-f:1600725448529658711'];
```
josteink: This global test-variable is only used inside one test. Putting it inside that test would IMO make the test easier to read.

test/test.tools.js
```diff
@@ -45,3 +47,40 @@ describe("Monkeypatching", () => {
...
+    const mockThreadData = data[2][7];
```
josteink: Same for this test-variable.

josteink:
I've looked over the changes so far, and I'm very happy with the overall structure of everything. And I really like that you've added test-data and tests for the new sceanrios.

If you say this is ready for proper review, I'll also conduct some functional testing, and if that goes through, I'll be happy to merge.

Thanks for hanging in there all this time. It's really appreciated!

josteink:
Review summary:

* [x] Code looks good
* [x] Code has tests
* [x] Code actually runs and does what it says

Thanks for hanging out in there. Great job! 

Thanks for the patch. It's really appreciated. Consider this merged!

Issue #48, opened 2014-04-17
Email data with conversation view off


ghost:
Hi there,

In #43 and #44 we dealt with email status, and the need to differentiate what's fully visible / partially visible / hidden / deleted in a conversation.

But I had not thought about the case when _conversation view_ is _off_ (via [settings](https://mail.google.com/mail/u/0/?zx=fnkwu8b0urgl#settings/general)), in which each email is seen in separate views. But even then, metadata include the whole thread and what the user sees is not what is actually handled.

I don't remember if it's been mentioned before, but a `is_in_current_view` email property would be nice.

Or maybe an alternative way of parsing emails (like `api.get.current_email_data()`).

josteink:
This issue has had no activity for years. Assuming no longer a problem. Closing.

PR #482, opened 2018-05-31
Allow to include a custom inbox query as control variable.

Navega:
This is related to: https://github.com/KartikTalwar/gmail.js/issues/48

The problem here is that I receive wrong emails in my extension when I select them from inbox tab.

Navega, software developer at Appfluence Inc.


josteink:
Looks good to me. Consider it merged!

Issue #484, opened 2018-05-31
open_mail is never triggered on opening an email from inbox mode on new Gmail UI in v0.7.3.


thinksysprem:
**gmail.js version: 0.7.3**

inside the `gmail.observe.on('load')` handler, I have this code:

`gmail.observe.on('load', function name() {
            var userProfile = _getUserProfile();

            _postMessage({
                type: "GMAIL_INIT",
                payload: {
                    user_profile: userProfile
                }
            });

            gmail.observe.before('send_message', beforeSendMessageHandler);
            gmail.observe.after('send_message', aftersendMessageHandler);
            gmail.observe.on("open_email", openMailHandler);
            closeMailHandler(); // gmail lib does not have a even for this and register is not working
            gmail.observe.on("compose", composeMailHandler);
            gmail.observe.on("save_draft", saveDraftHandler);
            gmail.observe.on("recipient_change", recipientChangeHandler);
            checkForOpenAtLoad();
        });`

All handlers are registered without a problem, Proof:
![open_email_xhr_registration_on_load](https://user-images.githubusercontent.com/18140689/40784447-4e514906-6503-11e8-8387-0a94dc93a30d.png)

**But on opening the email from inbox mode, the handler is never triggered.**

**Chrome version: Version 66.0.3359.181 (Official Build) (64-bit)**

**Other Active extensions:**
1. Google Docs
2. Google Sheets

Am I doing something wrong?

josteink:
That sounds very likely.

Gmail.js has a pretty big feature-set, and far from all of it is supported on new Gmail.

Work has been done to ensure some basic stuff works (like not crashing, getting email meta-data, handling multiple concurrent Gmail.js-versions, etc), but lots of stuff remains.

One of the things where probably least have been done is events.

If you need to get an event working, I suspect the quickest/easiest way to get there is by providing a PR yourself.

In the mean time... Any reason you use the `open_email` instead of the `view_thread` and `view_email` events (which should be working)?

thinksysprem:
`open_email` gave us the 'id' of the opened thread directly while `view_thread` gave us the jQuery element which would require scrapping to find out the message id from the element. We require the id to get the data about the email asynchronously via `gmail.get.email_data_async`. 

![image](https://user-images.githubusercontent.com/18140689/40955764-a83b32c0-68a9-11e8-885d-257ce661af0e.png)

[...]

PR #486, opened 2018-06-07
add read/unread actions

hankypanky666:
add read/unread actions for new gmail ui and dataLayer

example: 
```
gmail.observe.before(
                'unread',
                (oldIds, paramsUrl, paramsBody, newEmailIds, newThreadIds) => {
                    startForcedSync();
                }
            );
```


josteink:
Wow. *That* is all the changes which is needed?

Nice! Guess that means we did some solid work on the first PR :)

I’ll have to test this before merging, but if it works I have no further objections.

josteink:
I just tested this, and as far as I can see, the `unread` event works fine. However the `read` event doesn't trigger at all, at least not from the Inbox.

I've used the following code for verification:

````javascript
gmail.observe.before(
                'unread',
                (...args) => {
                    console.log("UNREAD!!!", args);
                }
            );
gmail.observe.before(
                'read',
                (...args) => {
                    console.log("READ!!!", args);
                }
            );
````

Have you tested the `read`-event yourself? How did you make it work?

I'm getting some weird `read`-event notifications when navigating to a an already read email from the inbox, and I get them every time.

I'll have to consider that test failed for now.

That means we have at least 2 options:

* fix the `read`-event to trigger when expected, and just as importantly, NOT trigger for unrelated events.
* remove the support for the `read`-event, only merging support for `unread` for now.

What do you think?

josteink:
OK. I can verify that these events works as advertised.

If we merge this, we will then also fix https://github.com/KartikTalwar/gmail.js/issues/484 in the process :)

I'm OK with merging this now, as is. Is that OK with you too, or are there other changes you would like to make first?

hankypanky666:
ok, merging) as there will be time, I will try to make cheker for 'read' action, thanks!

PR #487, opened 2018-06-08
add checker for second type actions

hankypanky666:
hi! Now I explored what we have two type actions, simple `is_first_type_action ` and difficult `is_second_type_action ` (ex: mark as read, star and etc...).  
What you think about this? `("^u^us" : "read",)`
and this `("^ss_so^ss_sr^ss_sp^ss_sb^ss_sg^ss_cr^ss_co^ss_cy^ss_cg^ss_cb^ss_cp" : "star",)`


josteink:
Hey there.

Thanks for another great patch. You're almost single-handedly fixing all "new Gmail"-bugs of importance here :)

That said, as I do for all bigger chunks of code, I do have some comments on things I think may be structured differently, and have left comments about that.

Let me know what you think.

src/gmail.js
```diff
@@ -1137,11 +1149,16 @@ var Gmail = function(localJQuery) {
...
+            const action = api.tools.get_first_type_action(threadData);
 
             return action_map[action];
-        } else {
+        } else if (threadData && api.check.data.is_second_type_action(threadData)) {
+            const action = api.tools.get_second_type_action(threadData);
+
+            return action_map[action];
+        }
+        else {
```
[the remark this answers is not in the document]
hankypanky666: you are right, I was in a hurry to fix in my project ) I refactor and write tests tomorrow

hankypanky666:
Thanks for the review! Your library very helps us in our project and I decided to help you :)

josteink:
Lol! Looking at the diff now you've gone *crazy* on events and I'm absolutely loving that!

Your only problem with that is that you've gone crazy with adding events, but not so much with related tests... And that will always lead me to nitpicking your commits 😄 

Would it be possible to restructure the tests not to be about "first" or "second" kind of action... But rather about what kind of event they represent? Kinda like this mockup code here:

````javascript
// you'll want several of these. one per event!
const xhrData = testData.new_gmail_archive_action_body_params;

const api = new GMail();
const threads = api.tools.extract_from_graph(xhrData , api.check.data.is_thread);
const actionType = api.tools.check_event_type(threads[0]);

// ensure you have test for matching event-type!
assert.equal("archive", actionType);
````

This again can probably be generalized into a single test-helper:

````javascript
function testXhrEventParsing(xhrData, eventName) {
    const api = new GMail();
    const threads = api.tools.extract_from_graph(xhrData , api.check.data.is_thread);
    const actionType = api.tools.check_event_type(threads[0]);
    assert.equal(eventName, actionType);
}
````

Then you can create a series of tests for all events, simply by obtaining event-data and adding it to the `testData` structure:

````javascript
it("parses archived messages", () => {
    const xhrData = testData.new_gmail_archive_action_body_params;
    testXhrEventParsing(xhrData, "archive");
});

it("parses archived messages", () => {
    const xhrData = testData.new_gmail_read_action_body_params;
    testXhrEventParsing(xhrData, "read");
});

// etc
````

I really think that would raise the level of usefulness in these tests, not to mention provide a way more extensive coverage than we've had before.

How about it? 😃

**Comment by hankypanky666** [comment]
ok! this is a great idea)

**Comment by josteink** [comment]
Let me know when you have something you want me to review, and I'll be ready :)

**Comment by hankypanky666** [comment]
Hi! sorry, pre-release days))) In this Sunday/Monday I push new code)

File: test/test.tools.js [hunk 2]
```diff
@@ -50,37 +60,29 @@ describe("Monkeypatching", () => {
 
 describe("Test tools for parsing new gmail body_params", () => {
     const gmail = new Gmail();
-    const data =  JSON.parse(testData.new_gmail_archive_action_body_params);
-
-    it("get thread id", () => {
-        const thread = gmail.tools.get_thread_id(data);
```
josteink: This test is not related to event-parsing, but covers some primitives which event-parsing relies on, so I think this should stay.

test/test.tools.js
```diff
@@ -50,37 +60,29 @@ describe("Monkeypatching", () => {
...
 
-        assert.equal(thread, 'thread-f:1600724307680265309');
+    it("parses archived messages", () => {
+        const xhrData = testData.new_gmail_archive_action_body_params;
+        testXhrEventParsing(xhrData, "archive");
     });
-    it("get thread data", () => {
-        const mockThreadData = data[2][7];
```
josteink: Same with this one.

test/test.tools.js
```diff
@@ -50,37 +60,29 @@ describe("Monkeypatching", () => {
...
-        const threadData = gmail.tools.get_thread_data(data);
 
-        assert.deepEqual(threadData, mockThreadData);
+    it("parses deleted messages", () => {
+        const xhrData = testData.new_gmail_delete_action_body_params;
+        testXhrEventParsing(xhrData, "delete");
     });
-    it("get messages ids", () => {
-        const mockMessageIds = ['msg-f:1600724307680265309', 'msg-f:1600724938213937205', 'msg-f:1600725174437456906', 'msg-f:1600725319255992336', 'msg-f:1600725448529658711'];
-        const threadData = gmail.tools.get_thread_data(data);
```
josteink: And this one.

We just want to replace the tests which concretely cover action/events, and not other thread/message meta-data.

hankypanky666:
ok!) I fixed tests :)

test/test.tools.js
```diff
@@ -51,36 +61,53 @@ describe("Monkeypatching", () => {
 describe("Test tools for parsing new gmail body_params", () => {
     const gmail = new Gmail();
     const data =  JSON.parse(testData.new_gmail_archive_action_body_params);
+    const threads = gmail.tools.extract_from_graph(data, gmail.check.data.is_thread);
+    const threadData = threads.map(thread => gmail.tools.get_thread_data(thread))[0];
 
     it("get thread id", () => {
-        const thread = gmail.tools.get_thread_id(data);
+        const thread = gmail.tools.get_thread_id(threads[0]);
```
josteink: You've changed to parameter from `data` to `threads[0]`... Any particular reason for that?

test/test.tools.js
```diff
@@ -51,36 +61,53 @@ describe("Monkeypatching", () => {
...
 
-        assert.equal(thread, 'thread-f:1600724307680265309');
+        assert.equal(thread, 'thread-f:1603171109786600032');
     });
+
     it("get thread data", () => {
-        const mockThreadData = data[2][7];
-        const threadData = gmail.tools.get_thread_data(data);
+        const mockThreadData = threads[0][2][7];
+        const threadData = gmail.tools.get_thread_data(threads[0]);
```
josteink: Same here. Why are you changing these tests?

test/test.tools.js
```diff
@@ -51,36 +61,53 @@ describe("Monkeypatching", () => {
 describe("Test tools for parsing new gmail body_params", () => {
     const gmail = new Gmail();
     const data =  JSON.parse(testData.new_gmail_archive_action_body_params);
+    const threads = gmail.tools.extract_from_graph(data, gmail.check.data.is_thread);
+    const threadData = threads.map(thread => gmail.tools.get_thread_data(thread))[0];
```
josteink: This value is only used in a single test, and should be put inside that test for improved clarity.

test/test.tools.js
```diff
@@ -51,36 +61,53 @@ describe("Monkeypatching", () => {
...
+        const threadData = gmail.tools.get_thread_data(threads[0]);
+        console.log(threadData);
 
         assert.deepEqual(threadData, mockThreadData);
     });
+
     it("get messages ids", () => {
-        const mockMessageIds = ['msg-f:1600724307680265309', 'msg-f:1600724938213937205', 'msg-f:1600725174437456906', 'msg-f:1600725319255992336', 'msg-f:1600725448529658711'];
-        const threadData = gmail.tools.get_thread_data(data);
+        const mockMessageIds = ['msg-f:1603171109786600032', 'msg-f:1603245801543734539', 'msg-f:1603245862071354412', 'msg-f:1603246018478443087', 'msg-f:1603256094012730022', 'msg-f:1603256564311088576', 'msg-f:1603256665279246114', 'msg-f:1603256682384715664', 'msg-f:1603376719891477511', 'msg-f:1603376909485932601', 'msg-f:1603376994923202634', 'msg-f:1603380395240179639'];
```
josteink: I don't see why this test needed to be changed. Has there been changes to the test-data?

Also: What XHR test data are we actually working against here?

These changes are confusing me.

test/test.tools.js
```diff
@@ -51,36 +61,53 @@ describe("Monkeypatching", () => {
...
-        assert.equal(messagesIds.length, 5);
+        assert.equal(messagesIds.length, 12);
         assert.deepEqual(messagesIds, mockMessageIds);
     });
-    it("get action type", () => {
-        const threadData = gmail.tools.get_thread_data(data);
-        const action = gmail.tools.get_action_type(threadData);
 
-        assert.equal(action, "^a");
+    it("parses archived messages", () => {
```
josteink: All these changes down here is exactly what I'm looking for *on top of* the existing tests. This is good stuff.

This should allow us to fairly quickly build test-cases covering actual XHR data, so that we can confidently assess that we don't have any regression.

If you can keep everything down here, while reverting the changes to the existing tests, I'll be much happier.

test/test.tools.js
```diff
@@ -5,6 +5,16 @@ let gmail = new Gmail();
...
+function testXhrEventParsing(jsonXhrData, eventName) {
+    const api = new Gmail();
+    const xhrData = JSON.parse(jsonXhrData);
+
+    const threads = api.tools.extract_from_graph(xhrData, api.check.data.is_thread);
+    const actionType = api.tools.check_event_type(threads[0]);
+
+    assert.equal(eventName, actionType);
+};
+
```
josteink: This is a good test-helper. This looks good to me.

test/test.tools.js
hankypanky666: I changed data in `testdata-parser.js` and now we have clear data, the same as sending us gmail )

test/test.tools.js
josteink: This could probably be simplified to `const threadData = gmail.tools.get_thread_data(threads[0]);` or am I missing something?

test/test.tools.js
josteink: Right. In which case that's technically an improvement. Good good 😃 

That said, we're changing things in several places here, and as a result I'm struggling to keep full context of the code.

Wouldn't it make more sense to just modify the base test-data setup instead of every where it's used then? Like this?

````diff
-const data =  JSON.parse(testData.new_gmail_archive_action_body_params);
+const xhrData =  JSON.parse(testData.new_gmail_archive_action_body_params);
+const threads = gmail.tools.extract_from_graph(data, gmail.check.data.is_thread);
+const data = threads[0];
````

And then we can leave all the existing tests as is to minimize the diff and get focus on the "important" stuff?

test/test.tools.js
hankypanky666: this is your func)))

test/test.tools.js
hankypanky666: Yes, I changed test-data, now we have clear data before I used prepared data

test/test.tools.js
hankypanky666: maybe you are right I wanted to repeat logic from code)

test/test.tools.js
hankypanky666: I think prev test data did not meet the requirements we have good test data for other events now. 
In this line we prepare data
`const data =  JSON.parse(testData.new_gmail_archive_action_body_params);`
in this we prepare threads from clear gmail response
`const threads = gmail.tools.extract_from_graph(data, gmail.check.data.is_thread);`

Maybe in future, we need to refactor this code for another tool but now it looks clearly.

What do you think?)

test/test.tools.js
hankypanky666: yes, we may add to `test-data` new event data(see new_gmail_delete_action_body_params and etc... for example) and write short tests for this data now

josteink:
I’ll have to take a deeper look tomorrow and try to clear up any confusion I still have. 

Thanks for hanging out in there 🙂

josteink:
I've done some functional testing, and yes I can confirm all new events works. That's very nice work!

I also really appreciate how you took the time to implement much better testing than we used to have for this stuff. That's the kind of stuff which will allow us to work more confidently with the codebase in the future.

I realize now though (and this is entirely my fault!) that you've added your new tests and test-setup to `test.tools.js` and not `test.parsing.js`. That's obviously wrong, but again, all my fault since I steered you in that direction.

I'll clean that up myself, and I'll merge everything to master when I'm done. Thanks for another good PR!

hankypanky666:
thanks! I will continue to add actions as far as possible

Issue #481, opened 2018-05-30
The selected emails are wrong.


Navega:
Hey,

I detected an issue related to this: https://github.com/KartikTalwar/gmail.js/issues/417

I had to fix removing your code here: https://github.com/appfluence/gmail.js/commit/86f4e0992ae2b2a6f7cc082ce965356e0da871aa

Could you tell me the reason to include these lines?

`// tentative fix for https://github.com/KartikTalwar/gmail.js/issues/417
                 if (api.check.is_google_apps_user()) {
                     url += "&search=" + "inbox";
                 } else {
                     url += "&search=" + "mbox";
               }`

josteink:
There's lots of discussion around these lines here in various issues, including #417.

For me personally, this is code I never use myself, so I really have no opinion on these, expect from what I'm told from people using and testing them (on several accounts, gmail, gsuite, old, new, etc).

And until now, this has supposedly been the best way to get reliable results for this function.

If that's no longer the case, feel free to issue a PR on these matters, but please note we will need to have some test-data/reports that shows that it works consistently across all the above-mentioned configurations, because we don't want to risk breaking stuff for *others*.

If you create such a PR, please CC @ other people from those issues so they too can assist testing.

josteink:
> reason because I didn't a PR with changes, I can't justify that this is the best solution

I can completely sympathize.

> For now, I will continue working in other branch

That's absolutely fair.

But what you *could* do if you want to stay on mainline `gmail.js` and seamlessly get bug-fixes and improvements without having to maintain your own fork is this...

How about you extend the `api.get.visible_emails()` functions with an optional search/query-parameter which takes precedence over the specific code-section you've pasted above:

````javascript
            } else {
                if (customInboxQuery) {
                    url += "&search=" + customInboxQuery;
                }
                // tentative fix for https://github.com/KartikTalwar/gmail.js/issues/417
                else if (api.check.is_google_apps_user()) {
                    url += "&search=" + "inbox";
                } else {
                    url += "&search=" + "mbox";
                }
            }
`````

That way, you get to control the behaviour you want in your extension without having the huge burden of proving that is the one correct way everyone should do it.

It's not *perfectly clean*, but I'd like to avoid fragmentation in the gmail.js ecosystem as much as possible, and would consider that a decent compromise.

If you think that sounds OK, feel free to issue a (much smaller) PR for that change and I will review it quickly.

Navega:
[...]

Issue #417, opened 2017-08-08
gmail.get.selected_emails_data() returns wrong emails


scottwrobinson:
I recently noticed when `gmail.get.selected_emails_data()` is called (and emails are selected) the wrong emails are returned. When viewing the "Primary" inbox tab, it seems to be returning emails from either "Social" or "Promotions" of the same index number.

Tried to debug a little and it seems `gmail.helper.get.visible_emails_pre()` is returning this URL:

`https://mail.google.com/mail/u/0/?ui=2&ik=b92f69409f&rid=undefined&view=tl&num=120&rt=1&start=NaN&sstart=NaN&search=inbox`

Requests to this URL seems to be returning emails from the "Social" and "Promotions" tabs.

Could the `rid=undefined` and `start=NaN&sstart=NaN` be causing the problem? Let me know what info you need from me to help resolve the issue. Thanks!

scottwrobinson:
Quick update:

After spying on Gmail's XHR requests I tried some different parameters in the request URL and it seems that I get the correct emails with `search=mbox` instead of `search=inbox`. The parameters `rid`, `start`, and `sstart` didn't seem to have an affect on returning the correct emails.

Could the fix be as easy as replacing

`url += "&search=" + "inbox";`

with

`url += "&search=" + "mbox";`

at [this line](https://github.com/KartikTalwar/gmail.js/blob/f532bae2841111216bbe221a2334e2f40cbecc14/src/gmail.js#L1740), or is it more complicated than that?

Thanks!

josteink:
> at this line, or is it more complicated than that?

It might be that simple, or it might be more complicated :)

It's hard to tell without doing some more testing. I for one don't use these different modes with "Primary", "Social" and what not. I just use it in the "classic" mode where you have an inbox and that is all.

We'd need to make sure your changes works properly across these different setups, and preferably also across regular Gmail accounts vs Google Apps/GSuite accounts.

Right now I'm a little short on time, but if you can assist with testing, outline what you've tested and prepare a PR, it will make it significantly easier for me to get these changes through.

scottwrobinson:
I'll test as many setups as I can. Ideas on different configurations to test are welcome (aside from regular Gmail accounts, GSuite accounts, etc). Thanks!

josteink:
Some suggestions:

- Classic inbox vs "new" inbox. 
- Testing on page 2 or 3. 
- Testing on labels/folders or searches. 

I think that should be considered extensive enough.

[...]

PR #489, opened 2018-06-22
Added support for category tabs in api.get.selected_emails_data()

scottwrobinson:
When category tabs (see below) are present in the inbox and we call `gmail.get.selected_emails_data()`, the wrong emails are returned. This PR fixes that.

screen shot 2018-06-21 at 6 38 14 pm

`api.get.visible_emails()` returns the correct emails within `gmail.get.selected_emails_data()`, but `$("[gh='tl'] div[role='checkbox']")` selected checkboxes from all category tabs, and not just the current one, throwing off the index and thus which emails were returned from this method.

This PR may be related to #417 (which I opened a while back) and possibly #481.

I've tested this out in the old Gmail UI, new Gmail UI, and with/without category tabs. Let me know if you have any questions.

Thanks for all your work on gmail.js!

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -2186,7 +2186,15 @@ var Gmail = function(localJQuery) {
             if($("[gh='tl'] div[role='checkbox'][aria-checked='true']").length){
                 var email = null;
                 var emails = api.get.visible_emails(customInboxQuery);
-                $("[gh='tl'] div[role='checkbox']").each(function(index){
+
+                // Select the open tab if "Primary", "Social",
+                // and "Promotions" tabs are present
+                var emailPanel = $("[gh='tl']");
+                if (emailPanel.find("div div[role='tabpanel']").length) {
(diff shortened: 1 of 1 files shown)


josteink:
Hey Scott, welcome back and thanks for another patch!

I can reproduce the errors that you've mentioned.

Looking at the code, it seems like mostly reasonable changes and I don't object to anything w.r.t. the code-structure. This is also not really something we can unit-test meaningfully, so lack of tests isn't really a concern for me either.

But I did test this out manually, and unfortunately there still seems to be some issues: Basically it seems to *still* select the wrong emails (although different wrong emails).

Testing on new and old Gmail, I get what seems like a completely random mail instead of the selected one inside all but the primary category tab (while I used to get 1-off results before).

Could you take a look at the code on your end, and see if you can reproduce the same... Or possible issues for us getting different results?

josteink:
I haven’t seen any response to my questions or feedback on this one, so closing due to inactivity. 

Please reopen if you want to resume this PR.

PR #490, opened 2018-06-27
is_conversation_view() view not working is fixed.

shashikiran-im:
Fixed the issue of the method is_conversation_view() not working in new gmail UI. Now it works in both classic and new gmail interface.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -2567,22 +2567,29 @@ var Gmail = function(localJQuery) {
 
 
     api.check.is_conversation_view = function() {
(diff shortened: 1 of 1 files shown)


josteink:
Hey @shashikiran-im and thanks for the patch!

Any help getting support or new Gmail is really appreciated.

That said, the CI-build fails because the code-linter has some issues with your formatting, and we only merge green builds.

Before I take the time to verify and manually test your patch, could you please fix the linting errors?

shashikiran-im:
Will generate new pull request after fixing lint errors.

Issue #492, opened 2018-06-29
"gmail.get.displayed_email_data()" is not working when conversation is set "off"


shashikiran-im:
Hi,

Issue: 
When conversation view setting is set to "off" then the method "gmail.get.displayed_email_data()" is returning empty in both classic and new Gmail GUI.

**Analysis did so far:**
displayed_email_data() method internally invokes "get_displayed_email_data_for_single_email()" method and it fires below jquery statement which fails as HTML structure in both classic and new Gmail UI is missing email id in the element shown below,
jQuery:
$(".ii.gt .a3s.aXjCH." + message_class_id);

**New Gmail HTML Structure:**
<div id=":3z" class="ii gt" style="display:none">
<div id=":40" class="a3s aXjCH undefined"></div></div>  //It should have be "m163bbb042cc0db69" but it is having "undefined"

**Solution:** 
Now when email is opened in both classic and new Gmail GUI interface we can locate below element. 
So we can rely on below statement instead of above statement and it is working fine.
var displayed_email_element = $("div[data-legacy-message-id='" + message_class_id + "']");

<div class="adn ads" style="display:" data-message-id="163bbb042cc0db69" data-legacy-message-id="163bbb042cc0db69">

Kindly let me know if the above solution is feasible so that I will generate pull request.

Thanks,
Shashikiran

josteink:
> So we can rely on below statement instead of above statement and it is working fine.
var displayed_email_element = $("div[data-legacy-message-id='" + message_class_id + "']");

It is true that Google has recently added these attributes to assist extension compatibility in a standard way. As such, I absolutely support using them where possible.

And some places in the code-base we already do:

````
$ grep -i "data-legacy-" gmail.js
        const elem = document.querySelector("h2[data-legacy-thread-id]");
            element = $("div.adn[data-legacy-message-id='" + this.id + "']");
````

While expanding the usage of this attribute just for sake it sounds "risky", I think migrating to it for the sake of fixing bugs is absolutely reasonable.

Looking at the implementation for `get_displayed_email_data_for_single_email` it seems the only thing we're using this element for is to verify if a message with a given ID is actually exists and is visible/accessible via the DOM, in which case we add it to the list of returned email-data objects.

As such, I think it sounds like a perfectly reasonable bug-fix, and I can't really see any major risk associated with this change.

Thanks for the debugging! Do feel free to make a PR for this!

[...]

PR #497, opened 2018-07-04
#492 - gmail.get.displayed_email_data() is not working when conversation is set "off"

shashikiran-im:
1. Updated get_displayed_email_data_for_single_email method to fix the jquery selector not working issue in new gmail interface
2. Above fix inturn fix the issue gmail.check.is_conversation_view()

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -2529,10 +2529,10 @@ var Gmail = function(localJQuery) {
     var get_displayed_email_data_for_single_email = function(email_data) {
         var displayed_email_data = {};
         for (var id in email_data.threads) {
-            var message_class_id = "m"+id;
-            var displayed_email_element = $(".ii.gt .a3s.aXjCH." + message_class_id);
+            var message_class_id = id;
+            var displayed_email_element = document.querySelector("div[data-legacy-message-id='" + message_class_id + "']");
 
-            if (displayed_email_element.length > 0) {
+            if (displayed_email_element) {
                 var email = email_data.threads[id];
 
                 displayed_email_data.first_email = id;


src/gmail.js
```diff
@@ -2529,10 +2529,10 @@ var Gmail = function(localJQuery) {
     var get_displayed_email_data_for_single_email = function(email_data) {
         var displayed_email_data = {};
         for (var id in email_data.threads) {
-            var message_class_id = "m"+id;
-            var displayed_email_element = $(".ii.gt .a3s.aXjCH." + message_class_id);
+            var message_class_id = id;
+            var displayed_email_element = document.querySelector("div[data-legacy-message-id='" + message_class_id + "']");
```
josteink: This looks much better, but lets just drop the `message_class_id` variable all together:

* it's exactly the same as `id`
* it's not a message_class anymore

So yeah. Let's just drop it.

josteink:
Hey there and thanks for another patch!

That looks good, and exactly the kind of changes I had in mind. I just have one comment on the implementation, but apart from that, I'm happy about this PR.

If you can fix that, I'll try to get some testing done and before merging.

src/gmail.js
```diff
@@ -1127,6 +1127,7 @@ var Gmail = function(localJQuery) {
     };
 
     api.tools.check_event_type = function(threadObj) {
+        const label = "^x_";
         const action_map = {
```
josteink: This bit foes not belong in this PR.

src/gmail.js
```diff
@@ -1137,7 +1138,7 @@ var Gmail = function(localJQuery) {
             // ""            : "discard_draft",
             // ""            : "expand_categories",
             // ""            : "filter_messages_like_these",
-            // ""            : "label",
+            "^x_"            : "label",
             // "^io_im^imi": "mark_as_important",
```
josteink: This change does also not belong in this PR.

src/gmail.js
```diff
@@ -1166,7 +1167,12 @@ var Gmail = function(localJQuery) {
...
             const action = api.tools.get_action(threadData);
 
-            return action_map[action];
+            //Check if label is applied to email / existing email is moved to an label
+            if(action.startsWith(label) && api.check.data.is_first_type_action(threadData)) {
+                return action_map[label];
+            } else {
+                return action_map[action];
+            }
         } else {
```
josteink: Same with this change.

src/gmail.js
josteink: If you can fixup your PR to only include this bit, with the tiny review-comment I had, that would be great and I will merge.

PR #508, opened 2018-07-31
add "it" to api.tools.i18n

tosolini:
Add italian translation


josteink:
Looks good to me. Thanks for the patch!

PR #512, opened 2018-08-07
api.get.storage_info percent undefined

tosolini:
Sometime used var was undefined, this cause an error to percent variable, because used.replace cannot handle undefined. I put a dirty workaround to set percent as zero in this conditional situation.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -334,7 +334,11 @@ var Gmail = function(localJQuery) {
         var div = $(".md.mj").find("div")[0];
         var used = $(div).find("span")[0].text;
         var total = $(div).find("span")[1].text;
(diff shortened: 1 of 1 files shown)


josteink:
Hey and thanks for the PR.

We always appreciate patches to help out fix bugs, but for this one, I think we will need to think of a different approach.

Your solution is ignoring a failure and just returning *something* (which may or may not be right). If it doesn't matter if the result of the API-call is correct or not... Why even have that API call at all? And why would you call such a function?

I'd rather we try to find out why getting the values fail and fixing that, rather than outright ignoring the failure.

If you can investigate why our lookup fails (the real source of the failure), and other ways to obtain these numbers, I will be happy to merge a PR based on that kind of solution.

tosolini:
Ok, for now i cannot find the problem, maybe is google itself, i'll investigate further

tosolini:
with new gmail interface the problem is disappear, so no further actions needed.

PR #527, opened 2018-09-07
Bugfix/unread email

tosolini:
In the new version (2018) of Gmail the function api.get.unread_*_emails not get the correct value of unread items, because the new version don't report in the anchor value this number, but only the label.

Digging to the DOM we can see switching from text to title is the solution, also compatible with older version too.

The fix is also apply to:
unread_draft_emails
unread_spam_emails
unread_forum_emails
unread_update_emails
unread_promotion_emails
unread_social_emails

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -561,8 +561,8 @@ var Gmail = function(localJQuery) {
         var dom = $("div[role=navigation]").find("[title*='" + api.tools.i18n("inbox") + "']");
 
         if(dom.length > 0) {
-            if(dom[0].text.indexOf("(") !== -1) {
-                return parseInt(dom[0].text.split(":")[0].replace(/[^0-9]/g, ""));
+            if(dom[0].title.indexOf("(") !== -1) {
+                return parseInt(dom[0].title.split(":")[0].replace(/[^0-9]/g, ""));
             }
         }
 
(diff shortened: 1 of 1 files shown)


josteink:
Hey @tosolini and thanks for the patch!

Looking at this it mostly looks good and we definitely want this merged.

There's a 2 things we need to sort out though:

* You've submitted 2 different functional changes in 1 PR (localization AND implementation). It's customary to submit 1 PR per area changed, so that the changes can be discussed and merged individually. Could you submit the localization changes in a different PR?
* There's also some conflicts. Can you manually fix these, or base your patches on latest code in master to avoid the conflicts?

If you do that, review gets less messy, and we get a nice clean merge, which is easy to track later.

Let me know what you think :)

tosolini:
I have tried, but i don't know how to do it. This is made automatically from github and there isn't a selector or something else to choose the right PR. I also tried to delete it, but come back every time. Sorry my knowledge about github is very basic or less.

josteink:
OK. Fair enough. Not everyone is a Git or Github expert. Excemptions can be made :)

You can try to resolve the merge-conflicts directly here on Github: https://github.com/KartikTalwar/gmail.js/pull/527/conflicts

Note: This is a web-editor, so all you need to do is fix the areas which have conflicting commits and ensure whatever remains is the correct code only.

Let me know if you have any more issues.

tosolini:
Thanks.

josteink:
Looking at the total change now, it seems you've removed the Italian translations. Was that intentional?

tosolini:
Travis-ci seems don't like the code i removed, don't know why.. is the same of PR #508.

josteink:
Change looks mostly good to me. I can try to clean it up on my end. Thanks for your patience.

For less issues with PRs, always pull in latest from master and create a dedicated branch for your patch. You should get less issues that way :)

tosolini:
Yeah, you're right. Sorry this time i've used only web version and everything goes wrong :(

PR #533, opened 2018-09-28
exclude messageId with bump from predicate check

abrahamdio:
When user sends an email, there are two XHR calls that would be made by the new gmail ui. My thought is there is a time buffer (depending on the user preferences) for user to undo the email and Gmail needs to know whether to send the email or not from the server side. First call is made after the user clicks send, second call is made after X seconds when user does not undo the action. 

The first `/sync` call will return an object that includes the new messageId format (`msg-[a|f]:...`) and no details on the email data. 

The second `/sync` call will include email data along with the legacy messageId (`16xxxx....`). In the email data, there is a messageId that looks something like `msg-a:bump-r-3994200xxxx....`. This is probably to tell Gmail server to actually send the email (since user does not undo it). 

Small change, is to exclude `msg-f:bump-123...` or `msg-a-:bump-r-123...` from the predicate check function.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -982,8 +982,8 @@ var Gmail = function(localJQuery) {
     api.check.data.is_email_id = function(id) {
         return id
             && typeof id === "string"
-            && /^msg-[a|f]:/.test(id)
-            && id.indexOf('bump-') === -1;
+            && id.indexOf('bump-') === -1
+            && /^msg-[a|f]:/.test(id);
     };
 
     api.check.data.is_email = function(obj) {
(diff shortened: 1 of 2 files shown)


josteink:
Thanks for the PR!

The description is good and the changes seems reasonable. 

I’ll do some testing when I’m back at the office and merge if I have no further questions or comments.

abrahamdio:
@josteink I pushed some minor changes, that should make the code _a little bit_ better. Let me know if there is any questions.

josteink:
Sorry about the slow response. I've been moving and haven't had much time for anything else lately.

I've now done some testing, and from what I can tell it looks good to me.

Thanks for the patch, and again: Sorry about the delay in handling it.

PR #534, opened 2018-10-04
feat: add `dom.header()` method

Boffka:
Add the ability to get the header.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,7 @@ These methods return the DOM data itself
 - gmail.dom**.email_body()**
 - gmail.dom**.email_contents()**
 - gmail.dom**.get_left_sidebar_links()**
+- gmail.dom**.header()**
 - gmail.dom**.search_bar()**
 - gmail.dom**.toolbar()**
 - gmail.dom**.right_toolbar()**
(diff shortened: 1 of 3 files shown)


josteink:
Looks good to me. Tested and does what it says.

The existing JSDoc in the TypeScript definitions is a bit messy, and that kinda ruins your PR, but I've documented that in https://github.com/KartikTalwar/gmail.js/issues/535, and I won't hold that against you :smile: 

As for the implementation: Right now, you just return a jQuery object, and I guess that is OK. But is that all you imagine we're going to use this method for now and in the future?

If you have other future use-cases, where just using jQuery won't be sufficient (like dom.email and dom.compose) it's probably better to create a type for that now rather than later.

If you foresee no such other use-cases... I'm OK to merge this as is.

Boffka:
@josteink thanks for feedback.
Why did I find this method attractive? 
Header is an important element. If the community will be use this method in their extensions , and together we will follow the Gmail layout changes - we'll get a bit of stability and benefits as opposed to the use of jQuery directly. 
This is my initial request and not the last. I will be glad to take part in the development of this library.
Cheers!

josteink:
Fair enough. Looks good to me. Consider it merged. 

Thanks for the patch!

PR #548, opened 2019-01-03
Fixed the issue 546

shashikiran-im:
is_conversation_view() was not working:
Root cause of the issue:
1. is_conversation_view() is always returning true irrespective of conversation view is set to ON / OFF.
2. Gmail had changed the response in there GLOBALS variables.

Resolution:
And found that it is reflected in the array api.tracker.globals[24] has the value "7164" then conversation view is ON if it is not present in the array then it is OFF.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -2790,8 +2790,8 @@ var Gmail = function(localJQuery) {
     api.check.is_conversation_view = function() {
         if( api.check.is_new_data_layer() ) {
             var conversation_flag = undefined;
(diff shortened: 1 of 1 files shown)


josteink:
Testing on New Gmail (which Google is force-transitioning all accounts to), this seems to work as expected for me. I can't test on old Gmail, but I don't consider compatibility with old Gmail critical at this point.

So all in all, LGTM!

Thanks for the PR! Consider it merged!

PR #556, opened 2019-02-24
fix #555 - wrong typescript typings for GmailDomCompose

### LINKED ISSUE #555: Wrong typescript typings for GmailDomCompose
version 0.9.1

currently in **GmailDomCompose** _recipients_, _to_, _cc_ and _bcc_ methods described like:
```
recipients(options?: { type: string, flat: boolean }): GmailEmailAddress[];
to(): string;
cc(): string;
bcc(): string;
```
but actually we have the next picture

selection_011

as you see _recipients_ returns object, not array of GmailEmailAddress
and _to_, _cc_ and _bcc_ returns jQuery object

bagbee:


src/gmail.d.ts
```diff
@@ -495,19 +501,19 @@ declare type GmailDomCompose = {
        options.type  string  to, cc, or bcc to check a specific one
        options.flat  boolean if true will just return an array of all recipients instead of splitting out into to, cc, and bcc
     */
-    recipients(options?: { type: string, flat: boolean }): GmailEmailAddress[];
+    recipients(options?: { type: 'to' | 'cc' | 'bcc', flat: boolean }): GmailDomComposeRecipients | string[];
```
josteink: All changes up to this point looks good to me, as they're mostly documenting what we're doing. This is good.

src/gmail.d.ts
```diff
@@ -495,19 +501,19 @@ declare type GmailDomCompose = {
...
     /**
       Retrieve the current 'cc' recipients
      */
-    cc(): string;
+    cc(): string[];
     /**
       Retrieve the current 'bcc' recipients
      */
-    bcc(): string;
+    bcc(): string[];
```
josteink: These changes don't document what we're currently doing.but rather document what they *will* be with the changes to Gmail.js itself further down.

While so far not really documented, this *is* a public API, and I wouldn't want to change it needlessly. This could cause unexpected breakage for people upgrading.

Couldn't we just instead document what it currently does, i.e. document this as returning jQuery? To me, that sounds a lot safer.

src/gmail.js
```diff
@@ -3184,7 +3184,7 @@ var Gmail = function(localJQuery) {
             var name_selector = options.type ? "[name=" + options.type + "]" : "";
 
             // determine an array of all emails specified for To, CC and BCC and extract addresses into an object for the callback
-            var recipients = options.flat ? [] : {};
+            var recipients = options.flat ? [] : { to: [], cc: [], bcc: [] };
             this.$el.find(".GS input[type=hidden]"+name_selector).each(function(idx, recipient ){
```
josteink: I'm completely OK with this change. This makes using the results safer.

src/gmail.js
```diff
@@ -3199,22 +3199,22 @@ var Gmail = function(localJQuery) {
...
         },
 
         /**
            Retrieve the current "bcc" recipients
         */
-        bcc: function(bcc) {
-            return this.dom("bcc").val(bcc);
+        bcc: function() {
+            return this.dom("bcc").siblings('.vR').find('input').map((i, el) => el.value).get();
         },
```
josteink: These however.... I'd rather not have these as they fundamentally alter behaviour people may depend on.

josteink:
I've left some comments on this PR which I think is worth discussing.

Could you please take a look and let me know what you think?

src/gmail.js
bagbee: yes, people may depend on
but this methods don't work correctly and returns jquery object with empty textarea which is useless

we can change this method and keep the previous behavior, for example:
```javascript
bcc: function(arrayOfStrings) {
    return arrayOfStrings
        ? this.dom("bcc").siblings('.vR').find('input').map((i, el) => el.value).get()
        : this.dom("bcc");
}
```

josteink: I've done some basic testing, and I have to admit the current implementation is not good. But your updated implementation only *retrieves* existing values and won't let a caller replace it, so it lacks a property this function used to have. So that's not all good either.

I certainly think this is something we should look into fixing properly, but I don't think this PR is the right place to do that.

Let's fix the type-annotations to match the implementation first, and *then* let's fix the implementation to act properly.

Sounds OK?

bagbee: yes, will update PR

bagbee: done

josteink:
LGTM. Thanks for helping out!

PR #559, opened 2019-03-26
Fix for undefined opener

yevhene:
You can often meet in tests `window.opener === undefined`. But currently it is checked only for `window.opener !== null` and my tests is failing.


josteink:
Looks fair enough to me.

@KartikTalwar: Any objections?

josteink:
With no objections registered, I'm just going to merge this.

Thanks for the patch, and sorry for the delay. Feel free to send more patches in the future!

josteink:
Now available in 0.9.2.

PR #567, opened 2019-04-24
Scheduling

bagbee:
Added support for send scheduled message event

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,8 +210,8 @@ gmail.observe.on("load", function(){
   - **`move_label`** - When a conversation(s) is moved to a label folder
   - **`save_draft`** - When a draft is saved
   - **`discard_draft`** - When a draft is dicarded
-  - **`send_message`** - When a message is sent
-  - **`send_scheduled_message`** - When a scheduled message is sent
+  - **`send_message`** - When a message is sent (except scheduled messages)
+  - **`send_scheduled_message`** - When a message is scheduled for sending (but not actually sent)
   - **`expand_categories`** - When a category is expanded from the left nav sidebar
   - **`restore_message_in_thread`** - When a deleted message is restored inside a thread
   - **`delete_label`** - When a label is deleted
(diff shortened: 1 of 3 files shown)


josteink:
Hey there and thanks for another PR.

I haven't *tested* the code in question, but apart from that everything else looks clean.

One question though: What's the correlation between `send_message` and `send_scheduled_message`? If I send a scheduled message which event(s) should I expect, and in what order?

Due to the nature of the code in question, I can't tell without experimenting, so this really needs to be spelled out somewhere. It would IMO be beneficial to add to the documentation.

Could you add a little extra to cover that topic?

bagbee:
Hey

There is no connection between these events.

`send_message` will work only for usual sending.
`send_scheduled_message` will work only for scheduled sending.

I decided to separate these two events, since the scheduling is not the actual sending.

Please let me know what do you think about it.

josteink:
Do you know if you schedule something 5 minutes from now... And keep the window open.

Will `send_message` fire 5 minutes later, or is that done completely in the dark?

bagbee:
when you scheduled a message the compose window automatically closes and the message goes into the scheduled label, without the possibility of editing (this is not a draft), user can only cancel sending

[video with scheduling flow](https://github.com/KartikTalwar/gmail.js/files/3114080/screencast-mail.google.com-2019.04.24-23-41-48.webm.zip)

josteink:
Looks legit. Thanks for putting in the effort in demonstrating it.

Looking at it that way, I would probably like to phrase it somewhat differently in the documentation:

- **`send_message`** - When a message is sent (except scheduled messages)
- **`send_scheduled_message`** - When a message is scheduled for sending (but not actually sent)

Unless I misunderstood how things work, I think that might be more descriptive and understandable for people reading the documentation and wanting to know what these events represents.

Could you touch those bits up too?

bagbee:
Done

josteink:
Great stuff. Thanks for the PR. 

Consider it merged!

bagbee:
@josteink hi,

just found out that there are no PR changes in npm repository
https://registry.npmjs.org/gmail-js/-/gmail-js-0.9.5.tgz

can you clarify it?

thanks

josteink:
That seems to be correct.

We need to bump the version and update the changelog, and then we can publish a new version with the changes from your PR. 

Sorry about the delay.

josteink:
0.9.6 is out now. Thanks for your PRs!

PR #570, opened 2019-06-11
 bugfixed: All apis of getting unread emails don't work, always return 0; 

hzb:


hzb:
@josteink Hello, does this have issues?

josteink:
Sorry for the slow reply!

Looking at the diff, this looks much better. 

Unfortunately I’m on vacation and won’t be able to actually test these changes for at least 2 weeks. 

If you remind me then, I’ll do my best to get this tested, expedited and merged quickly!

hzb:
Take your time, have a nice vacation.

josteink:
And *there* I'm back at the office, ready to do some testing! Whee!

I've only tested on English so far, but running a simple test though, I'm not sure I'm getting what's expected:

````javascript
// gmail.get.unread_emails()
{
  drafts: 1,
  forum: 0,
  inbox: NaN, // <-- this doesn't look correct
  promotions: 0,
  social: 0,
  spam: 318,
  update: 0
}
````

Is this the same results you are getting?

josteink:
This kind of of code is repeated lots of times, and will need to handle `0` gracefully.

````javascript
        var dom = $("div[role=navigation]").find("[title*='" + api.tools.i18n("inbox") + "']");
        if(dom.length > 0) {
            if(dom[0].title.indexOf(api.tools.i18n("inbox")) !== -1) {
                return parseInt(dom[0].attributes['aria-label'].value.replace(/[^0-9]/g, ""));
            }
        }

        return 0;
````

I propose we create a new function to identify node, and extract number from it (falling back to `0`), which we re-use for all different mail-boxes.

Does that sound good?

hzb:
From my PR, you can see that I just update some codes that will work fine for me, I think you don't need this PR, I will close it, thanks.

josteink:
You misunderstand. I definitely want this PR, but I was just discussing the code firm before merging. 

That’s both common and useful.

josteink:
Hey @hzb.

Can you please re-open this so we can fix the final issues and get this merged?

josteink:
Since @hzb did not want to come back and fix this, I took his code and committed it myself.

This issue is now fixed as per https://github.com/KartikTalwar/gmail.js/commit/7eb6d96df7eb92b15396ef5d06cd9e8831007d70 and will be in the upcoming 0.9.7-release.

PR #577, opened 2019-07-08
Attachment URLs not working for non-primary gmail accounts

ek23:
The attachment URL only works for the user's primary email address.  The user string /u/# must be inserted for non-primary email accounts

Both URLs work fine
https://mail.google.com/mail/?ui=2...
https://mail.google.com/mail/**u/0**?ui=2...

Only second URL works
https://mail.google.com/mail/?ui=2...
https://mail.google.com/mail/**u/1**?ui=2...

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1238,13 +1238,20 @@ var Gmail = function(localJQuery) {
                     attachment_id: item["1"]["2"],
                     name: data["3"],
                     type: data["4"],
-                    url: data["2"],
+                    url: api.tools.check_fd_attachment_url(data["2"]),
                     size: Number.parseInt(data["5"])
                 });
             }
         }
 
         return res;
+    };
+
+    api.tools.check_fd_attachment_url = function(url) {
+        var userAccountUrlPart = api.tracker.globals[7];
+        if (url.indexOf(userAccountUrlPart) < 0) {
+            url = url.replace('/mail/?', userAccountUrlPart + '?');
+        }
     };
 
     api.tools.parse_fd_request_html_payload = function(fd_email) {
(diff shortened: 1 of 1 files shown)


src/gmail.js
```diff
@@ -1247,6 +1247,13 @@ var Gmail = function(localJQuery) {
         return res;
     };
 
+    api.tools.check_fd_attachment_url = function(url) {
+        var userAccountUrlPart = api.tracker.globals[7];
+        if (url.indexOf(userAccountUrlPart) < 0) {
```
josteink: Should we handle the possibility of url being null or empty here?

It may not be needed, but in case of code running inside hijacked XHR requests, I'd rather be safe than sorry.

josteink:
Thanks for yet another good patch! Keep them coming 😄 

I'll be happy to have this merged, once review is done.

ek23:
That makes sense to me.  Is this what you are looking for?

src/gmail.js
```diff
@@ -1247,6 +1247,13 @@ var Gmail = function(localJQuery) {
         return res;
     };
 
+    api.tools.check_fd_attachment_url = function(url) {
+        var userAccountUrlPart = api.tracker.globals[7];
+        if (url && userAccountUrlPart && url.indexOf(userAccountUrlPart) < 0) {
+            url = url.replace('/mail/?', userAccountUrlPart + '?');
+        }
+    };
+
```
josteink: You should probably return the ‘url’ variable to the caller too, but otherwise yes, this looks good to me.

ek23:
Good point.  This should have been a quick one

PR #578, opened 2019-07-08
api.dom.email should use new id to populate element

ek23:
This api.dom.email constructor is still setting $el based on the legacy id, but the new id format will be passed in.  I tried to keep backwards compatibility here, but if we can ignore the legacy ids then we can just change the two lines that directly reference the legacy ids

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -3442,17 +3442,28 @@ var Gmail = function(localJQuery) {
         }
 
         if (typeof element === "string") {
-            this.id = element;
-            element = $("div.adn[data-legacy-message-id='" + this.id + "']");
+            if (element.substring(0,3) == 'msg') {
+                this.id = element;
+                element = $("div.adn[data-message-id='#" + this.id + "']");
+            } else {
+                this.legacy_email_id = element;
+                element = $("div.adn[data-legacy-message-id='" + this.legacy_email_id + "']");
+            }   
         } else {
             element = $(element);
         }
 
         if (!element || (!element.hasClass("adn"))) api.tools.error("api.dom.email called with invalid element/id");
 
-        this.$el = element;
-        if (!this.id) {
-            this.id = this.$el.data("legacyMessageId");
+        this.$el = element;        
+        if (!this.legacy_email_id) {
+            this.legacy_email_id = this.$el.data("legacyMessageId");
+        }
+        if (!this.id) {            
+            let email_id = this.$el.data("messageId");
+            if (email_id) {
+                this.id = email_id.substring(1);   
+            }         
         }
 
         return this;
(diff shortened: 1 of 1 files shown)


josteink:
Hey there and thanks for yet another patch!

Improving API ergonomics is a good thing, and I really support that. As such I'm happy about the intent of this PR.

> if we can ignore the legacy ids then we can just change the two lines that directly reference the legacy ids

That might break compatibility with existing extensions relying on `.id` referencing a legacy ID (not to mention enterprise-customers still on "old" gmail which has no new-style email-id!).

Maintaining compatibility is very important, so I'd rather not do that unless I know 100% for it to be safe. I hope that's OK.

Which leaves us with the code... I have a few comments and suggestions you may want to consider.

src/gmail.js
```diff
@@ -3442,17 +3442,28 @@ var Gmail = function(localJQuery) {
...
-            this.id = element;
-            element = $("div.adn[data-legacy-message-id='" + this.id + "']");
+            if (element.substring(0,3) === 'msg') {
+                this.id = element;
+                element = $("div.adn[data-message-id='#" + this.id + "']");
+            } else {
+                this.legacy_email_id = element;
+                element = $("div.adn[data-legacy-message-id='" + this.legacy_email_id + "']");
+            }   
         } else {
```
josteink: Consider using the existing functions in the `api.helper.get`-namespace.

Both `api.helper.get.legacy_email_id()` and `api.helper.get.new_email_id()` cover what you are trying to do here, in what may be a more reusable/simple way. These are designed for maximum ergonomics and have pretty good test-coverage, so you should be able to rely on them acting properly (and consistently with the rest of the code-base).

For now I advice using `api.helper.get.legacy_email_id()` to maximize backwards compatibility.

Sounds good?

src/gmail.js
```diff
@@ -3442,17 +3442,28 @@ var Gmail = function(localJQuery) {
...
-            this.id = this.$el.data("legacyMessageId");
+        this.$el = element;        
+        if (!this.legacy_email_id) {
+            this.legacy_email_id = this.$el.data("legacyMessageId");
+        }
+        if (!this.id) {            
+            let email_id = this.$el.data("messageId");
+            if (email_id) {
+                this.id = email_id.substring(1);   
+            }
```
josteink: From what I can tell, this change breaks compatibility by changing what the `.id`-property represents.

I'll be happy to add new-style ID properties, but only in addition to the existing `.id`-property which for compatibility-reasons I'd like to remain as it is.

Is that OK with you?

ek23:
It looks like get.new_email_id and get.legacy_email_id were not around when I coded this.  I am going to close this PR and revisit my changes to see if they are even necessary any more.  Thanks for the feedback.

PR #580, opened 2019-07-09
Parse sent email

ek23:
In parse_request_payload, handles 
TODO: parse 'email' for contents, and provide a better structured object

I didn't add any tests yet, but I think I could handle creating them with a little direction.  I have the unparsed json ready to add to test.parsing.js.


josteink:
Thanks for yet another patch.

This one looks really nice and useful and I know lots of people have been requesting this kind of stuff.

That said, before merging I'll have to do some testing, and we will also have to discuss how to handle compatibility with existing code in a clean way.

Sounds good?

ek23:
Sounds good.  We have had this one in production for a couple of months it has been working well.  I can help out setting up some of the tests, just let me know.

josteink:
Ok so I've tested and this looks really good. I definitely want this merged.

Only issue we have is really that `send_message` is already somewhat cludgy to handle in a compatible manner. In my extension I already have code like this:

````typescript
    private async onMessageSent(url: string, body: string, data: any, xhr: XMLHttpRequest): Promise<void> {
        // data has 2 possible structures:
        // - old gmail: .composeid is reference to triggering compose
        // - new gmail: ["1"] contains the actual new style gmail email-id.
        const composeId: string = data.composeid || data["1"];
````

I expect other people have done the same to extract the email-id in case of new Gmail as well. So while it might look ugly, I think the new email-format should also include a `"1"` property which holds the email-id, just for the sake of compatibility. Can you add that?

Apart from that I see two more things I'd like to get done here, before we consider this feature complete:

* Like you mention: test-cases. Help is definitely wanted. Can you push some as part of this PR? (Add some test-data to [testdata-parser.js](https://github.com/KartikTalwar/gmail.js/blob/master/test/testdata-parser.js) and add some test-cases to [test.parsing.js](https://github.com/KartikTalwar/gmail.js/blob/master/test/test.parsing.js), and just follow the pattern found in either `Response-parsing` or `Attachment-parsing` category.

* Typescript-definitions. If this is outside your comfort-zone, I can probably get this done after the PR is merged.

Sounds good? Let me know what you think!

josteink:
Hey @ek23 ! Still around?

I'd love to get this merged, but we still have a little bit of work to do 😄

ek23:
Yep, I was on vacation last week, so things are backed up this week.  I'll give the Typescript defs a try as well.

josteink:
Sounds great. Thanks!

ek23:
Gave it all a shot in one go.  Let me know if anything doesn't look right or needs to be expanded.  I wasn't too sure about the proper name of the TS object, so I just stuck it with the similar GmailNewEmailData.

josteink:
That looks great. Thanks! The TypeScript annotation looks good as it is too.

*Ideally* it could be directly coupled to the event, to avoid a need for manual declaration in the event-handler, but I'm honestly not sure how to do that with such "stringly-typed" events. If I find out, that can be an overall improvement later on.

All in all, I think this is a great PR. Consider it merged!

Issue #584, opened 2019-08-02
How is gmail.new.get.* a replacement for gmail.get.visible_emails()?


mikob:
Hi, thanks for the library! Let's say I need to programmatically select some emails on the users currently visible list of emails. `gmail.get.visible_emails` seemed like the right thing but I see it's deprecated and I'm advised to use `gmail.new.get.*` instead. 

I don't see which of these is a replacement for that method though:
`email_data`
`email_id`
`thread_data`
`thread_id`

none of these return the list of emails as far as I can tell. Perhaps I'm misunderstanding, how exactly are those a replacement? 

How would you advice getting the list of emails so I can select them. Should I use `gmail.dom.inbox_content()` and then find the rows that have checkboxes?

I will be happy to contribute to the library and help maintain, as I am working on a project that could potentially use this!

josteink:
> How is gmail.new.get.* a replacement for gmail.get.visible_emails()?

Good question! It isn't. Yet.

> none of these return the list of emails as far as I can tell

Correct.

> how exactly are those a replacement?

`api.new.get.email_data()` and `api.new.get.thread_data()` replace `api.get.email_data()` in the sense that `api.get.email_data()` will stop working, and those hopefully wont.

All other variations of `api.get.*_email_data()` are ultimately based on `api.get.email_data()` and will eventually stop working too.

What that ultimately means is that we've managed to provide decent replacements for `api.get.email_data()`, but not its contextual siblings.

> How would you advice getting the list of emails so I can select them. 

Use some jQuery/DOM selector magic to extract the new style IDs which represents visible emails, selected emails (or whatever) and use the new-style API to get data about those.

> Should I use gmail.dom.inbox_content() and then find the rows that have checkboxes?

If that works, great. Maybe you can inspect its code and see if there's room for optimizations when those two actions are combined too?

> I will be happy to contribute to the library and help maintain

Sounds good. If so, PRs are definitely welcome.

This is a open-source library and any functionality currently present in this library has been provided by contributors like yourself scratching your own itch.

I scratched my own itch, which was the general reverse-engineering required for `api.new.get.email_data()` and `api.new.get.thread_data()`, which was no small task let me tell you.

If you can base your code on those, and extend the API with new methods like `api.new.get.visible_email_data()` or `api.new.get.selected_email_data()` that would be great and I'll be glad to review and merge those PRs.

mikob:
[...]

PR #585, opened 2019-08-05
Get email messages, dom elements and data

mikob:
First stab at two new api calls. Slightly related to #584

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ const gmail = new GmailFactory.Gmail() as Gmail;
 - [gmail.get**.new.email_data()**](#gmailnewgetemail_dataidentifier)
 - [gmail.get**.new.thread_id()**](#gmailnewgetthread_id)
 - [gmail.get**.new.thread_data()**](#gmailnewgetthread_dataidentifier)
+- [gmail.get**.new.message_data()**](#gmailnewgetmessage_dataidentifier)
 
 - [gmail.get**.email_subject()**](#gmailgetemail_subject)
 - [gmail.get**.compose_ids()**](#gmailgetcompose_ids)
@@ -423,6 +424,28 @@ the first invocation returned `null`.
 }
 ```
 
+#### gmail.new.get.message_data(identifier)
+
+Returns basic data for all the messages currently visible in the messages view. Taken from the DOM.
+
+```json
+[
+{
+    "summary": "Hey Bill",
+    "from": {
+        "email": "joe@gmail.com",
+        "name": "Joe",
+    },
+},
+{
+    "summary": "The best of Gmail, wherever you are",
+    "from": {
+        "name": "Gmail",
+        "email": "mail-noreply@google.com"
+    },
+},
+]
+```
 #### gmail.get.email_source(identifier=undefined)
 
 Deprecated function. Will be removed. Migrate to
--- a/README.md
+++ b/README.md
@@ -95,7 +95,6 @@ const gmail = new GmailFactory.Gmail() as Gmail;
 - [gmail.get**.new.email_data()**](#gmailnewgetemail_dataidentifier)
 - [gmail.get**.new.thread_id()**](#gmailnewgetthread_id)
 - [gmail.get**.new.thread_data()**](#gmailnewgetthread_dataidentifier)
-- [gmail.get**.new.message_data()**](#gmailnewgetmessage_dataidentifier)
 
 - [gmail.get**.email_subject()**](#gmailgetemail_subject)
 - [gmail.get**.compose_ids()**](#gmailgetcompose_ids)
@@ -238,7 +237,7 @@ These methods return the DOM data itself
 
 - gmail.dom**.inboxes()**
 - gmail.dom**.inbox_content()**
-- gmail.dom**.visible_messages()**
+- [gmail.dom**.visible_messages()**](#gmaildomvisiblemessages)
 - gmail.dom**.email_subject()**
 - gmail.dom**.email_body()**
 - gmail.dom**.email_contents()**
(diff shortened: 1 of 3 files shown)


josteink:
Hey there, @mikob. Thanks for the patch!

First of all: Sorry about the slow response.

Just looking at the code (that means I haven't actually *tested* any of this yet), I have a few comments and questions:

* `api.new.get.message_data()` is *DOM* based, but is positioned in a namespace usually reserved for data-based operations.
* I can't see a way to reverse-trace data from `api.new.get.message_data()` back to individual DOM-nodes, or threads. There is no `.$el`-property nor a `.threadId` property. Was this a deliberate omission? 

I did a quick test in gmail, and from what I can tell, you should be able to extract thread-IDs from the nodes:

````javascript
        const msgEles = api.dom.visible_messages();
        const ret = [];
        for (const msgEle of msgEles) {
            const idNode = msgEle.querySelector("span[data-thread-id]");
            const threadId = idNode.dataset.threadId;
            // "#thread-f:1638756560099919527|msg-f:1638756560099919527"
            const legacyThreadId = idNode.dataset.legacyThreadId;
            // "16be0868ae18c6a7"
````

I really think you should add these, along with the element property.

And if you add those, what that message will return is .... a list of visible email-threads, based exclusively on DOM-data!

So how about we drop/inline `api.dom.visible_emails()`, and just have one function which does *everything*. And that function could be called  `api.dom.visible_emails()`?

I think API-wise that would work much better. Opinions?

**Comment by mikob** [comment]
> api.new.get.message_data() is DOM based, but is positioned in a namespace usually reserved for data-based operations.

OK, I thought the `dom.*` methods were for calls that returned DOM data, not just operated with it. Btw, why do we care if the API works with the DOM as opposed to other means?

Yes, I can combine both methods into api.dom.visible_emails() and add the `$el` and `threadId` props.

**Comment by josteink** [comment]
> OK, I thought the dom.* methods were for calls that returned DOM data, not just operated with it. 

There's several methods which work on DOM-based content and allows you to interact with it, `api.dom.compose` probably being the most common in use.

>  Btw, why do we care if the API works with the DOM as opposed to other means?

I think it matters.

It highlights to an extension-developer if the APIs they invoke depends on the user actually being navigated to a particular part of Gmail, with the associated page DOM.

If we don't signal this somehow, developers may expect these methods to work in contexts which they obviously not will. And IMO that's not a good API-design.

**Comment by josteink** [comment]
> OK, I thought the dom.* methods were for calls that returned DOM data, not just operated with it. 

There's several methods which work on DOM-based content and allows you to interact with it, `api.dom.compose` probably being the most common in use.

>  Btw, why do we care if the API works with the DOM as opposed to other means?

I think it matters.

It highlights to an extension-developer if the APIs they invoke depends on the user actually being navigated to a particular part of Gmail, with the associated page DOM.

If we don't signal this somehow, developers may expect these methods to work in contexts which they obviously not will. And IMO that's not a good API-design.

> Yes, I can combine both methods into api.dom.visible_emails() and add the $el and threadId props.

Great. Make it so!

If you could also update the documentation, not to mention add this new method to the typescript annotations (`gmail.d.ts.`) that would be great. Let me know if you need help with these.

**Comment by mikob** [comment]
@josteink how's this?

**Comment by josteink** [comment]
Hey there. Sorry about the slow response.

This looks pretty good! I still have a few tiny nitpicks, but really, after that I'll be happy :)

* the comment `// legacy: `does not match the content. The comment represents the value found by `idNode.dataset.threadId` and has little to do with "legacy" IDs as used elsewhere in GmailJS. As such, in its current form it's can be a cause of confusion. Can you replace `legacy: ` with for instance `expected value: ` ?
* We should apply the same kind of filtering to this `thread_id` as we did in the `api.compose.thread_id()`-patch. The value returned from this function *should* be directly usable in `api.new.get.thread_data()`! At this point it may be worth extracting the thread-id cleanup-logic into its own function?
* We should be explicit about the difference between legacy IDs and "new-style" IDs in our objects. Please create a dedicated property for `legacyMessageId`, and leave `thread_id` empty if none can be found there.

If you can make these changes, I promise this will get merged without much further delay :smile:

**Comment by mikob** [comment]
@josteink how's this?

File: src/gmail.js [hunk 1]
```diff
@@ -3291,6 +3306,40 @@ var Gmail = function(localJQuery) {
...
+                // example value: #thread-f:1638756560099919527|msg-f:1638756560099919527"
+                const idNode = msgEle.querySelector("span[data-thread-id]");
+                ret.push({
+                    from: {
+                        name: nameAndEmail.attr('name'),
+                        email: nameAndEmail.attr('email'),
+                    },
+                    summary: linkAndSubject[0].innerText,
+                    thread_id: api.helper.clean_thread_id(idNode.dataset.threadId || ""),
+                    legacyMessageId:  idNode.dataset.legacyMessageId,
```
josteink: We've used the identifier `legacy_email_id` other places in the code, so for consistency it would be nice to use here as well.

src/gmail.d.ts
```diff
@@ -487,6 +487,16 @@ declare type GmailDomComposeLookup =
...
 
+interface GmailMessageRow {
+   summary: string;
+   from: {
+      name: string,
+      email: string,
+   };
+   $el: JQuery;
+   thread_id: string;
+}
```
josteink: Missing `legacy_email_id` property :)

README.md
```diff
@@ -1059,6 +1060,33 @@ gmail.observe.on('compose_email_select', function(match) {
...
+[
+{
+    "summary": "Hey Bill",
+    "from": {
+        "email": "joe@gmail.com",
+        "name": "Joe",
+    },
+    "$el": tr#:9b.zA.zE.inboxsdk__thread_row,
+    "thread_id": "#thread-f:1634069952006597945"
+},
```
josteink: Also missing `legacy_email_id` in the documentation example.

josteink:
I've added one final nitpick, and if you fix that one too, I promise that is final.

Thanks for your patience and thanks for a great patch! I'll be happy to bump the version and make a new release for this one :smiley:

src/gmail.js
josteink: I see you've probably chosen this identifier due to me being imprecise in my previous comment, and I apologize for that.

But really, if we just make it `legacy_email_id` everything else will be good :smile:

josteink:
You know what... The "improper" `legacyMessageId` vs `legacy_email_id` thing was my partially bad. I'll pull in your changes and fix that up myself :)

Again: Thanks for the patch!

Issue #558, opened 2019-03-19
api.new.get.email_data is giving null. 


anilreddyr7:
Earlier I was using api.get.email_data to get email data. If I see newer gmail.js, it is suggested to use new.get.email_data as get.email_data is a deprecated Gmail API. How to use this function? If I invoke new.get.email_data instead of get.email_data, it is giving me **null** .Can you please help me out how we can get the data using new method? Do we have any sample code snippet?

josteink:
Then you APIs are definitely not document that that’s something we should look into fixing that let me see if I can make some quick examples sometime this week.  Then you APIs are definitely not document. That’s something we should look into fixing. 

Let me see if I can make some quick examples sometime later this week.

wnateg:
For me, it seemed to work more often, if I checked for the data periodically. I didn't check the code to verify there was any evidence to support this, but it worked with trial and error.

`
		var i = 0, interval = setInterval(function() {
			var dataNew = gmail.new.get.email_data(msgId);
			if (typeof dataNew === undefined || ! dataNew) {
				console.log(i, dataNew);
				return (++i) > 60 ? clearInterval(interval) : true;
			}
			var pixelId = (dataNew.content_html.match(/\/([\w]+)\.gif/i) || []).pop();

			if (typeof pixelId !== undefined) {
			}
			clearInterval(interval);
		}, 500);`

josteink:
With the new APIs we don’t actuslly “get” the email, but rather gmailjs prefetches and caches *all* emails gmail processes. When you request an email, you will only get it if it is found in the cache. 

For any random email Id, it will probably fail. For an email on screen it should succeed. 

Hope this helps.

anilreddyr7:
> With the new APIs we don’t actually “get” the email, but rather gmailjs intercepts and caches _all_ emails gmail processes. When you request an email, you will only get it if it is found in the cache.
> 
> For any random email Id, it will probably fail. For an email on screen it should succeed.
> 
> Hope this helps.

Thanks for the update. It is already 2019, but still old method gives proper email data.  I'm not leaning towards new method. Because I'm worried about intercepting gmail process and getting email data.  Will Google definitely block old gmail request to fetch email data?

josteink:
> Thanks for the update. It is already 2019, but still old method gives proper email data.

While testing New Gmail while it was in closed beta, any kind of XHR on our part seemed to fail, and thus the all the `get.email_data()` too methods. 

[...]

PR #587, opened 2019-08-24
Add Start Reply

kstubs:
Support for clicking Reply button in email

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -3776,6 +3776,25 @@ var Gmail = function(localJQuery) {
     };
 
     /**
+     *  Show a compose window
+     * @returns {boolean}
+     */
+    api.compose.start_reply = function() {
+
+        //The compose button
+        var composeEl = $(".ams.bkH")[0];
+
+        if(composeEl) {
+            var click = document.createEvent("MouseEvents");
+            click.initEvent( "click", true, false );
+            composeEl.dispatchEvent(click);
+
+            return true;
+        }
+        return false;
+    };
+
+    /**
      * Shadow API commands specifically made to interact with old gmail.
      * (And in the future we can either remove "regular"  api.get or replace it with something else)
      */


josteink:

Looking at the code itself, apart from using plain DOM-objects, it looks mostly like the `send()` implementation we have on `api.dom.compose`-instances: https://github.com/KartikTalwar/gmail.js/blob/50291f74cf2ed8dff243c573eec1b1710cd6ec8c/src/gmail.js#L3480-L3485

One key difference which I see though is that this seems to be implemented globally. Was that intentional?

Is that going to be a source of errors/limitations? Could there be cases where we would rather want to reply to a specific `api.dom.email`-instance? It also seems to reply to the *first* email it finds with a reply-button (as opposed to the last), and I don't think that is what people are going to expect.

It might make more sense to add this method to the `api.dom.email`-object, and maybe even return the resulting `api.dom.compose` once invoked?

What do you think?

Apart from that, once we settle *where* this could should be and how it should work, we will probably also want this added to the documentation and typescript type-annotations. For now though, that should probably wait.

josteink:
Hey @kstubs.

Did you read my reply? Do you have any opinion on it?

I'm not going to merge this as is, so if you want this added, you will need to provide me with some feedback :)

kstubs:
Hi Jostein,
Sorry for the late reply but I've just been pulled away to some other
things in last 1-2 months.  I'm not abandoning this pull-request, but just
not able to get back to it with proper and necessary time to make the
changes (which I really appreciate, your excellent feedback on this).
So, not dropped, not forgotten, but can we leave it as is for now?

Karl Stubsjoen
(602) 845-0006

josteink:
Fair enough. Just checking in :+1:

kstubs:
Getting back into this Jostein but having issues.  I researched this issue
#558
... and left my comments there.  Basically, email_data() returning Null.

Thanks!

Karl Stubsjoen
(602) 845-0006

kstubs:
Wow!  Just looking at recent issues in Github and see # 589 "cache missing
first couple messages" and realize this is exactly the bug I am
experiencing.  I can work around this for now!

Karl Stubsjoen
(602) 845-0006

josteink:
Agreed about #589. My plan is to get a PR supposedly fixing that merged as quickly as possible.
So let's just assume that's not a problem, and move on to *your* PR!

Looking at the current state of the code, I can't see any updates based on [my review so far](https://github.com/KartikTalwar/gmail.js/pull/587#issuecomment-524730256)... Or am I mistaken?

I'd really like for those things to get better fleshed out, or at least discussed before considering merging this. A public API is something one should be careful about introducing after all :smile:

josteink:
@kstubs Your issue should now be fixed in a new and shiny `1.0.0` release. Please read the changelog :)

josteink:
This PR hasn't been active or touched in *years*. Closing. Let me know if you're still interested in getting this merged.

PR #597, opened 2019-11-13
Use top window as xhr window

bagbee:
I can argue that in the new gmail UI I observe two different cases:
- the first and working case in which the `before_send_message` even works is the case when `top` window is used as a xhr window.
- the second and non-working case when `window_opener` is used as a xhr window.

The second case is reproduced very rarely and not stably. I managed to get it on my g-suite account.

I tested using only the `top` window as the xhr window, and events worked fine, even `before_send_message` event.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1709,27 +1709,30 @@ var Gmail = function(localJQuery) {
     };
 
     api.helper.get_xhr_window = function() {
-        var js_frame = null;
+        // var js_frame = null;
+        // if (top.document.getElementById("js_frame")){
+        //     js_frame = top.document.getElementById("js_frame");
+        // } else if (window_opener) {
+        //     js_frame = window_opener.top.document.getElementById("js_frame");
+        // }
+        // if (!js_frame){
+        //     if (window_opener) {
+        //         js_frame = window_opener.top;
+        //     } else {
+        //         js_frame = top;
+        //     }
+        // }
+        // var win;
+        // if (js_frame.contentDocument) {
+        //     win = js_frame.contentDocument.defaultView;
+        // } else {
+        //     win = js_frame;
+        // }
+        // return win;
 
-        if (top.document.getElementById("js_frame")){
-            js_frame = top.document.getElementById("js_frame");
-        } else if (window_opener) {
-            js_frame = window_opener.top.document.getElementById("js_frame");
-        }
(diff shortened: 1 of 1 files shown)


josteink:

That looks like a great simplification. My question is if it isn't too great 😄 

From what I can tell, this change is focused around *new Gmail only*, right?

I'm not sure how, but for the extension I'm maintaing I'm still seeing around 10% of my users are still on OLD gmail. If that is applicable to other extension developers too, we need to try to keep things compatible with old gmail too (at least for a while).

Do you think this current approach will work with old gmail?

If not, can you rewrite it to not alter behaviour for old gmail?

bagbee:
Hey,

I'm not sure that this will work for the Old Gmail UI, so we will consider that this solution applies only to the New Gmail UI.

As far as I know, google forcibly transferred all users to the New Gmail UI.

And I also have users who use the Old Gmail UI, but I rely on the `is_new_guy` method.

If we believe google, we can conclude that the method `is_new_guy` does not always work correctly or users use the basic HTML version of Gmail.

josteink:
> As far as I know, google forcibly transferred all users to the New Gmail UI.

I have anonymised telemetry in my extension and it reports 10% of my users are still on old gmail. 

I suspect these are big enterprise/Gsuite accounts, and I don’t want to break those. 

For now old and new must remain supported.

josteink:
> if we believe google, we can conclude that the method is_new_gui does not always work correctly or users use the basic HTML version of Gmail.

That function is implemented by a hard spec provided to me by Google, and it has always been 100% reliable in all cases I have encountered. 

I’d rather believe google has decided not to force-upgrade big customers who don’t want their systems changed.

bagbee:
Ok, then I will rewrite it to alter behaviour only for New Gmail UI.

Also what do you use for telemetry in your extension?

josteink:
> Ok, then I will rewrite it to alter behaviour only for New Gmail UI.

Great!

> Also what do you use for telemetry in your extension?

It's a custom/proprietary system built into the CRM-software my extension integrates with. So unfortunately, it's not anything I can share.

bagbee:
done

josteink:
Thanks! Those changes looks reasonable to me. I've done some testing, and from what I can tell, everything still works.

Consider it merged 😃

PR #598, opened 2019-11-20
GmailJS 493 - Issue is fixed

shashikiran-im:
Change Log,
1. Fixed the issue of apply label event not working when multiple labels are applied to single / multiple emails from inbox is fixed.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1012,7 +1012,7 @@ var Gmail = function(localJQuery) {
         return obj
             && obj["1"]
             && Array.isArray(obj["1"])
-            && obj["1"].length === 1
+            && obj["1"].length >= 1
             && typeof obj["1"]["0"] === 'string';
     };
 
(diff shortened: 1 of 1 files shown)


src/gmail.js
```diff
@@ -1176,8 +1177,14 @@ var Gmail = function(localJQuery) {
...
-            return action_map[action];
+            
+            //Check if label is applied to email / existing email is moved to an label
+            if(action.startsWith(apply_label) && api.check.data.is_first_type_action(threadData)) {
+                return action_map[apply_label];
+            } else {
+                return action_map[action];
+            }
+            
         } else {
```
josteink: We now have an `if` nested inside an `if`. I really don't like stuff like that. That's how spaghetti-code starts 😄 

Could you make this into a three-way `if`, `else if`, `else` instead? If you start with the `null`-condition, that would end up looking much cleaner.

josteink:
Hey there and thanks for yet another patch! Keep them coming 😄 

For this PR, I think the code looks mostly good, apart from 1 comment I've left requesting some changes.

What I'd also like to see before merging is a test-case which verifies that this code actually works. Could you please add that as well?

PR #607, opened 2020-04-09
visible_messages fix

mikob:


josteink:
Haven’t tested yet, but otherwise it looks good to me. 

Will merge if/when I have time after Easter.

josteink:
Sorry about the slow response.

I just tested, and on my main inbox, where 50 emails are shown, I get this response:

Screenshot-2020 04 22-07 43 03

That is 75 emails are returned, instead of the 50 I expected. That sounds like too much to me, right?

josteink:
Disregard the previous (now deleted) message. I tested `api.get.visible_emails()`, not `api.dom.visible_messages()`.

As far as I can tell the code works properly, so consider it merged :smile:

Issue #589, opened 2019-08-29
The threadCache is always missing the first couple threads in the Inbox


ek23:
It looks like Google has recently made a change to how it retrieves threads when first loading Gmail.  The first i/fd XHR call asks for thread number six in the inbox.  The later calls ask for emails further down in the folder.  There is never an i/fd call asking for the first five threads.  Those first five threads are actually returned with the request to https://mail.google.com/mail/u/0/data.  

Has anyone else noticed this.  Seems like gmail.js would need to intercept the results of that /data call or just use the dom exclusively for those first five threads.

josteink:
Hey there and thanks for the report!

Also: Sorry about the slow reply. Been really busy putting out fires elsewhere.

That said: I've always noticed that the first email (when loading straight into an email-view) is missing from the threadCache, but never considered it crucial enough to investigate further.

If now more emails are starting to miss from the threadCache, I guess that will lead to a perception about the new GmailJS Data-API not being sufficiently stable.

> use the dom exclusively for those first five threads.

DOM is sadly insufficient to get the data-granularity we want. If the DOM-data is sufficient, I would always advice to use the DOM-API directly since it's probably going to be more reliable, not to mention faster.

> Seems like gmail.js would need to intercept the results of that /data call

That's probably a better idea. I'll have to research how though: This call is not  a regular XHR call, but seems to be a regular web-request so the XHR interceptor wont work.

Finding out if this is possible, and if so exactly how, will probably take some research.

You already seems to have looked a bit into this. Would you care to dig a little more? :smile:

josteink:
Also: Can confirm the most recent items in the Inbox is unavailable in the threadCache. 

This will definitely go noticed by many users in many extension, and should IMO really be fixed.

josteink:
Doing some very basic testing, I just added the following patch to log all intercepted requests:

````diff
diff --git a/src/gmail.js b/src/gmail.js
index c9803f2..8d8a40d 100755
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1668,6 +1668,7 @@ var Gmail = function(localJQuery) {
                     var xhr = this;
                     this.onreadystatechange = function(progress) {
                         if (this.readyState === this.DONE) {
+                            console.log("INTERCEPTED REQUEST:" + this.xhrParams.url_raw);
                             if (progress.target.responseType === "" || progress.target.responseType === "text") {
[...]

PR #608, opened 2020-04-10
Fix 589 - extend xhr parsing and email cache

Fabi1Sc:
Content of this pull request :
------------------------------------

 - Fix issue #589 with a parsing of embedded JSON data (see more details below)
 - Extend XHR parsing data with a parsing of /bv request
 - Extend email cache : update cache and not only append new email in cache
 - Refactoring function name (parse_fd_email2, parse_fd_email)

Changes :
-------------

    Fix issue #589
    -------------------

    Important : content-script which launch injected script must be configured with "run_at": "document_start" 
    It's recommand to split injected script to have only gmail.js to load in first because size of injected script impact the loadind time.
    Gmail.js must be injected and loaded before Gmail load embedded data.

    api.tools.embedded_data_watcher override Gmail JS function "_GM_setData"
    There is 2 different parsing (like XHR) : parse_fd_embedded_json and parse_bv_embedded_json

    Extend XHR parsing data with a parsing of /bv request
    ------------------------------------------------------------------------

    Like /fd request parsing, implement of /bv request parsing.

    Extend email cache : update cache and not only append new email in cache
    -----------------------------------------------------------------------------------------------------

    Update function api.tools.cache_email_data

More details :
-------------

In Gmail, there are 2 methods to load thread and email data.
    1) embedded data
    2) XHR

When you load Gmail UI with default URL (like https://mail.google.com or https://mail.google.com) the default view displayed 
is the inbox view, in this case, sequence is STEP 1 and STEP 2A

If you load Gmail UI with personnalized URL (like https://mail.google.com/mail/u/0/#sent), the wiew display depend on the URL.
in this case, sequence is STEP 1 and STEP 2B.

STEP 1 :
--------

This step is mandatory because it's automatic when loading Gmail UI.

Thread and email data for INBOX (BUT NOT ONLY these datas) can partially be read in JSON embedded data from text/html content request.
IT IS NOT an XHR...

First Http request is on https://mail.google.com/mail/u/0/, the response is a document text/html.
In the body of HTML content, we can find an iframe.
Example :
<body jscontroller="hS6RLb" jsmodel="utMpr" jsaction="............">
<iframe id="embedded_data_iframe" src="https://mail.google.com/mail/u/0/data?sw=2&amp;token=%5B%22cftp%22........." class="invfr" aria-hidden="true"></iframe>

When iframe is loaded, the document is text/html and contain javascript function and JSON embedded data
Example :
<html><body><script nonce="5schwSe/q2lcOQe6UONGRw">var AF_initDataKeys = []
; var AF_dataServiceRequests = {}; var AF_initDataChunkQueue = []; var AF_initDataCallback; var AF_initDataInitializeCallback;
[.........]
top._GM_setData({"a6jdv":[["sils",null,"{\"1\":{\"1\":[{\"2\":\"thread-f:166042295594152783

The SECOND JS function in document "top._GM_setData({"a6jdv":...})" load ALL thread and email data requisite to display the inbox_content view.
If inbox_content view display 50 thread or email also JSON contains the 50 thread or email.
It is not detail datas requisite to display the view_email or view_thread.
We can see later in description than is like "/sync/u/0/bv" request.

The FOURTH JS function "top._GM_setData({"Cl6csf":...})" contains detail data of the FIVE FIRST thread or email.
The FIVE FIRST thread or email data are never load with XHR, only with embedded data.
We can see later in description than is like "/sync/u/0/fd" request.

STEP 2:
-------

The sequence of STEP2 is different, it'a serie of XHR request.
Different xhr request with the path "/sync/u/0" are observe.

My comprehension of XHR fonctionnal use :

https://mail.google.com/sync/u/0/i/bv   => contains global data requisite to display a list of thread and email list
https://mail.google.com/sync/u/0/i/fd   => contains detail data requisite to display a specific thread and email 
https://mail.google.com/sync/u/0/i/s    => contains data only if there is new datas for thread and email already load
https://mail.google.com/sync/u/0/el2    => //TODO, need investigation

STEP 2A :
--------

This step concern the case of loading GMail UI on inbox view.
Global thread and email datas and five thread and email detail datas are loaded in STEP1.

Below an example of sequence :

T0 is the moment when user make one action to load inbox Gmail and nothing else.

For the case that inbox_content view display 50 threads in "Primary" tab, 2 thread in "Promotions" tab and 1 thread in "Notification" tab
=> T1 : T0 + 10s
X requests /fd to read data for the last 45 thread of "Primary" tab
1 request /bv with no thread data
=>20 seconds passed

=> T2 : T1 + 2min
1 request /s with the 25 first message (need a verification...) all tabs concerned
1 request /bv with no thread data
2 request /s with no thread data
1 request /fd with the 3 thread of "Promotions" tab and  "Notification" tab

=> T3 : T2 + 2min
1 request /s with no thread data

=> T4 : T3 + 2min
1 request /s with no thread data

STEP 2B :
--------

This step concern the case of loading GMail UI on other view than inbox.
Global thread and email datas and five thread and email detail datas are loaded in STEP1 but there not used.

T0 is the moment when user make one action to load inbox Gmail and nothing else
For the case that view display 50 threads:
T0 :
1 request /bv with the 50 thread and email data
X requests /fd to read data for the 50 thread


josteink:
Hey there. 

I’d just like to acknowledge that I’ve seen this PR, and skimmed the description. 

As a maintainer, it looks like one of the better PRs I’ve seen, and it’s _really_ appreciated! Absolutely solid work. 

That said, I’m kinda short on time right now, and due to the fundamental nature of these changes, I’d like to review and test them thoroughly before I merge. I hope that’s understandable. 

So please accept a little delay before getting feedback. It should hopefully be less than a week.

Fabi1Sc:
No problem, take your time ;)

josteink:
OK, so I haven't actually gone over the *code* yet, but I've tested these changes in the context of the boilerplate-project. From what I can tell it works really well!

One thing I notice is that some of these new emails seems to lack the legacy ID property:
Screenshot-2020 04 17-16 58 39

Did you notice that too? I don't really think is going to be problem, `smtp_id` is the important one IMO, but just an observation.

Also in the context of the boilerplate project, I wasn't able to reproduce any difference based on `document_start` vs `document_end`, but I'll take your word for it. It sounds timing-sensitive after all.

Now that I know that it works as it should, I'll try to get a look at the actual code as soon as possible and then get this merged :+1:

Fabi1Sc:
> One thing I notice is that some of these new emails seems to lack the legacy ID property:
> Screenshot-2020 04 17-16 58 39
> 
> Did you notice that too? I don't really think is going to be problem, `smtp_id` is the important one IMO, but just an observation.

No it's an oversight because I don't use this data `legacy_email_id`.
I commit a fix.

However i found a bug du to my code (on `smtp_id` for example).

```js
// only append email to cache if not already there.
if (thread.emails.filter(i => i.id === email.id).length === 0) {
    //console.log("push",email);
    thread.emails.push(email);
}
else
{
    let index = thread.emails.findIndex(i => i.id === email.id);
    //console.log("before:",thread.emails[index]);
    thread.emails[index] = email;
    //console.log("after:",thread.emails[index]);
}
```
There are less datas to read and store in bv request than in fd request.

If bv request is execute before and fd request after (classic case) it's good to update cache  because datas read on fd request are more detailled.

But I found a case where a bv request is execute after fd request, cache is updated and  it's bad because there is a data lost. 
The fix is simple, I think that email cache must be update only with detailled data from fd model data (xhr or embedded json). I think to make a patch this week end or monday.

josteink:
> I commit a fix.

I suddenly realized I really need this, so this is really great. 👍

> But I found a case where a bv request is execute after fd request, cache is updated and  it's bad because there is a data lost.

I guess ideally we should have a merge-algorithm of sorts when a cache needs to get updated?

To ensure we always end up with the richest data possible, no matter the loading order. 

> I think to make a patch this week end or monday.

No rush. Take your time. 🙂

Fabi1Sc:
> I guess ideally we should have a merge-algorithm of sorts when a cache needs to get updated?
> 
> To ensure we always end up with the richest data possible, no matter the loading order.
> 

Yes you right.
From what i saw, there is never less data in fd request compared  to bv request, so i thing algorithm is simple, we have just to update full data or not.

Fabi1Sc:
Hey

I push a patch to update email data cache depending on data source.

src/gmail.js
```diff
@@ -1885,15 +1906,18 @@ var Gmail = function(localJQuery) {
...
                 thread.emails.push(email);
             }
             else
             {
-                let index = thread.emails.findIndex(i => i.id === email.id);
-                //console.log("before:",thread.emails[index]);
-                thread.emails[index] = email;
-                //console.log("after:",thread.emails[index]);
+                // Only update cache with data source fd_request_payload and fd_embedded_json
+                if (data_source === "fd_request_payload" || data_source === "fd_embedded_json") {
```
josteink, line 1915: This could be expressed as an `else if` to reduce nesting.

src/gmail.js
```diff
@@ -1870,8 +1877,22 @@ var Gmail = function(localJQuery) {
...
-            c.emailLegacyIdCache[email.legacy_email_id] = email;
+            if (c.emailIdCache[email.id] === undefined) {
+                //console.log("ADD c.emailIdCache[email.id]",c.emailIdCache[email.id],email);
+                c.emailIdCache[email.id] = email;
+            }
+            else if (data_source === "fd_request_payload" || data_source === "fd_embedded_json") {
+                //console.log("UPDATE c.emailIdCache[email.id]",c.emailIdCache[email.id],email,data_source);
+                c.emailIdCache[email.id] = email;
+            }
+            if (c.emailLegacyIdCache[email.legacy_email_id] === undefined) {
```
josteink, line 1888: Is updating the main cache and the legacy_cache independently really the best thing to do? Now we need to duplicate certain logic based on source.

Wouldn't it better to first add email to main cache, and then afterwards, based on the current contents of the main cache, add it to the legacy-cache? Something like this:

````javascript
}

const cachedEmail = c.emailidCache[email.id];
c.emailLegacyIdCache[cachedEmail.legacy_email_id] = cachedEmail;
````

Or would that cause some sort of errors which is being handled by the currently proposed implementation?

File: src/gmail.js [hunk 3]
```diff
@@ -1870,8 +1877,22 @@ var Gmail = function(localJQuery) {
 
         for (let email of email_data) {
             // cache email directly on IDs
-            c.emailIdCache[email.id] = email;
-            c.emailLegacyIdCache[email.legacy_email_id] = email;
+            if (c.emailIdCache[email.id] === undefined) {
+                //console.log("ADD c.emailIdCache[email.id]",c.emailIdCache[email.id],email);
+                c.emailIdCache[email.id] = email;
+            }
+            else if (data_source === "fd_request_payload" || data_source === "fd_embedded_json") {
```
josteink, line 1884: If these two sources are the only sources which are going to be acted on, wouldn't it be better to just combine them into a boolean, like `isAuthorativeSource` or something?

I think it might make for fewer errors when calling this function, and easier to understand the semantic meaning of the parameter.

src/changes_notes.txt
```diff
@@ -0,0 +1,137 @@
+
+Content of this pull request :
+------------------------------
```
josteink, line 3: While I definitely appreciate the great description outlined in the PR, there's no need to add it as a separate file in the PR itself. :sweat_smile:

josteink:
One thing I see missing, which I would love to see to prevent regressions and future errors, is some test-data and test-cases for the new parser-functions.

While not super-great and not super-detailed, we *do* have some test-cases for the existing parsing code in `test.parsing.js` and `testdata-parser.js`. Could you add some test-cases for the new code in there?

src/changes_notes.txt
Fabi1Sc: I agree, I remove this file

src/gmail.js
Fabi1Sc: I agree, I combined the 2 data source with a boolean.
Code is clearer.

src/gmail.js
Fabi1Sc: You right, emailIdCache and emailLegacyIdCache must be load or update together, it's more secured. I changed that.

src/gmail.js
Fabi1Sc: I agree, done

Fabi1Sc:
> One thing I see missing, which I would love to see to prevent regressions and future errors, is some test-data and test-cases for the new parser-functions.
> 
> While not super-great and not super-detailed, we _do_ have some test-cases for the existing parsing code in `test.parsing.js` and `testdata-parser.js`. Could you add some test-cases for the new code in there?

I haven't really looked and I missed these tests.
ok to add test-case for new parsers in few days.

josteink:
> I haven't really looked and I missed these tests. ok to add test-case for new parsers in few days.

Sounds good. Looking forward to it!

Fabi1Sc:
For information on master branch one test failed :

```
_New Gmail event-triggering
Gmail.js encountered an error trying to parse sent message! TypeError: Cannot read property '7' of undefined
```

You did well to tell me to add test because actual test don't passe because of my code.

josteink:
> Also in the context of the boilerplate project, I wasn't able to reproduce any difference based on `document_start` vs `document_end`, but I'll take your word for it. It sounds timing-sensitive after all.

After testing this version on my production extension, I can indeed confirm this being required.

Turns out the boilerplate project is too small/fast (on my machine at least) to cause issues w.r.t. intercepting data, but for my much larger production extension I needed to split out gmailJs into its own bundle and load that early, to ensure things work 100%.

That's *kinda* a breaking change, in that people might be required to also update their extensions when updating to this latest version.

And as such... I'm planning on making the next release the long awaited 1.0 release! :tada: 

And it couldn't have happened without your help. Thanks for staying in there with this great PR :smiley:

Fabi1Sc:
> That's _kinda_ a breaking change, in that people might be required to also update their extensions when updating to this latest version.
> 

I'm not completely sure that's a a breaking change.
If you don't change anything in your extension, everything continue to be fine with new code in this PR. But if you want to exploit 100% of API features (including the fix #589), you have to update your extension with taking account of requirements.

josteink:
> I'm not completely sure that's a a breaking change. If you don't change anything in your extension, everything continue to be fine with new code in this PR. 

Fair enough. Anyway, making this a major version upgrade should signal to people that they can't upgrade blindly, so I still think bumping the version to 1.0 is the right thing to do.

josteink:
Any news here? Need any help getting the tests working?

Feel free to push a (broken) test-commit, and maybe I can help diagnose what's going wrong?

Fabi1Sc:
I just pushed changes with test case.
With test, I found and fxed 2 bugs with fd_embedded_json (content_html and attachment) 😀
Test case could be more detailled, I don't understad why but I have a problem with assert equal on html content

josteink:
That looks very nice indeed. Thanks for taking the time!

I'll take a deeper look at it tomorrow, and see if I can help out with the HTML test-cases. If we're lucky, it's something simple and easy to fix. :+1:

josteink:
Looking at the failing tests, it seems to be to be very specific for the emails under test.

From what I can tell all the failures are related to unicode-technicalities (composed vs non-composed apostrophes) and significant vs insignificant unicode-whitespace. Things which are easy to get wrong in almost any case.

Seeing as the HTML to be parsed properly apart from that, I did what you according to TDD should never do: reverse-engineered the test based on the test-data using `JSON.stringify(content_html)`.

This escaped the troublesome content and made all tests pass, and should at the same time provide us with protection against regressions in the future.

You can find the diff for the this change attached here: [html_fix.patch.txt](https://github.com/KartikTalwar/gmail.js/files/4543738/html_fix.patch.txt)

With that landed, I think we are getting pretty close to finally getting this merged! :smiley:

Fabi1Sc:
Thanks for your help.
I push fix.

josteink:
Tested the final code on two extensions, and it's looking great! :smiley: 

Thanks for this great patch. This will no doubt make Gmail.JS a better library for everyone, with less needs for workaround, and hopefully better compatibility with future Gmail versions when Google eventually completely turns off the old data-API we've been using in the past.

If you have any other contributions you want to make in the future, they are very much welcome!

And again: It's really been a long ride. Thanks for hanging out in there. Not everyone would :sunglasses: 

With that said: Consider it merged!

PR #616, opened 2020-07-01
Updated start_compose button selector

mikob:
it was changed


josteink:
Oh my.

Sorry about the delay. And you are absolutely right.

I'll get this merged and updated on npmjs.com pronto!

josteink:
Published as version 1.0.1.

Issue #620, opened 2020-10-19
Gmail.check.is_inside_email() seem to not work anymore


yevhene:
There are no items matching `.ii.gt .a3s.aXjCH` selector anymore on the page

josteink:
Hey there and thanks for reporting in.

I just did some quick testing in the browser and it seems you are correct. Just relaxing the selector slightly, down to to `".ii.gt .a3s"` instead seems to work better now.

Would you be willing to do some testing and see if that works better on your end too? Or if there are unexpected side-effects of such a change?

yevhene:
@josteink Work for me

PR #621, opened 2020-10-20
Fix Gmail.check.is_inside_email()

josteink:
This fixes https://github.com/KartikTalwar/gmail.js/issues/620.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -411,7 +411,7 @@ var Gmail = function(localJQuery) {
             return false;
         }
 
-        var items = $(".ii.gt .a3s.aXjCH");
+        var items = $(".ii.gt .a3s");
         var ids = [];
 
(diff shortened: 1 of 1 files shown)


josteink:
I've done a basic fix, and some basic-testing and it *looks right* to me, but that's about it.

But since I'm not using `gmail.check.is_inside_email()` in my production extension, so I don't have any extensive use-cases where getting it right is that important.

Could any of you guys who have expressed interest in this function in the past help test this change, and ensure we are not creating any regressions or false positives?

cc: @AlexBazer @jocstar @yevhene

yevhene:
I can confirm, it works for me on production perfectly

josteink:
I'll get this merged then!

josteink:
Fix now released in version `1.0.4`.

PR #622, opened 2020-10-21
Gmail.compose.start_compose fix

yevhene:
Gmail.compose.start_compose don't work now. Replacing of sophisticated trigger_mouse_click logic with just click solves the issue.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -2773,16 +2773,7 @@ var Gmail = function(localJQuery) {
     // dispatch mousedown and mouseup event on passed element
     api.helper.trigger_mouse_click = function(element) {
         if(element) {
-            //Trigger mouse down event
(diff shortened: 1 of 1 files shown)


josteink:
I'm pretty sure I wrote that complicated logic... And for a reason, although that reason escapes me now. If it has stopped working though, that definitely sucks.

As for your patch: It looks like a simple fix, and I'd be happy to merge, but only on one condition: that it has been tested in all supported browsers (Chrome & Firefox).

I'm not accepting any Chrome-only code into this library :smile:

r0642879659th (approved):

josteink:
Tested in both Chrome and Firefox just now. Whatever we had that work-around for before, it seems to be no longer needed. :+1:

Consider this merged. Sorry about the slow response!

PR #623, opened 2020-10-26
Wealthbox

yevhene:


josteink:
Hey there.

There was already a fix for your second commit in this PR: https://github.com/KartikTalwar/gmail.js/pull/621

Sorry for being slow at merging it, but it should now be available in `1.0.4`.

As for warnings, I want them there by default... So users of this library will actually know their code is at risk, and consider migrating to the newer APIs.

If you would like to send in a PR for muting those, I will only accept them if they are muted by setting a option/flag on the GmailJS instance like `gmail.DISABLE_DEPRECATION_WARNINGS` or something noisy developers will have to manually set up.

yevhene:
@josteink Thanks. Will switch on new version.
```
Gmail.new.get.email_data(Gmail.new.get.thread_id())
```
Seem to not work for me at all. I did't do extensive testing, will report if this behavior will be constantly reproducible.

josteink:
In new Gmail API thread_id and email_id is different. This is not the fault of GmailJS. 

You won't get email_data with thread_id, like you won't get thread_data with email_id. 

Use right get-method with the right id.

yevhene:
@josteink Accidentally created this PR in main repo)

PR #632, opened 2021-01-20
Correct grammar on README.md

ShivanKaul:
It took me several seconds to parse what was being said in that section of the README.

On a non-grammar note, this section recommends that the content script be configured to run at document start. However, the [linked example repo has the content script run at document end](https://github.com/josteink/gmailjs-node-boilerplate/blob/master/manifest.json#L16). Which one is correct?


josteink:
Start is correct. 

Thanks for the PR!

PR #637, opened 2021-02-13
vars used and total need to be turned into jquery objects

ibercovich:
vars used and total need to be turned into jquery objects and .text() needs to be used instead of .text

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -358,7 +358,7 @@ var Gmail = function(localJQuery) {
 
 
     api.get.storage_info = function() {
-        var div = document.querySelectorAll(".md.mj")[0].querySelectorAll("div")[0];
+        var div = document.querySelector(".md.mj div");
         var used = div.querySelectorAll("span")[0].textContent;
         var total = div.querySelectorAll("span")[1].textContent;
         var percent = parseFloat(used.replace(/[^0-9\.]/g, "")) * 100 / parseFloat(total.replace(/[^0-9\.]/g, ""));
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -359,8 +359,8 @@ var Gmail = function(localJQuery) {
 
     api.get.storage_info = function() {
         var div = document.querySelectorAll(".md.mj")[0].querySelectorAll("div")[0];
-        var used = div.querySelectorAll("span")[0].textContent
-        var total = div.querySelectorAll("span")[1].textContent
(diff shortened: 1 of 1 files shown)


josteink:
Hey there and thanks fir the patch!

I haven't tested this yet, but on the surface of it, it looks OK to me. 

That said, I'd like to move away from jQuery, and if we are to fix this code with a rewrite, I'd prefer for it to use more native DOM APIs and less jQuery. 

Do you think you can do that? 🙂

src/gmail.js
```diff
@@ -358,9 +358,9 @@ var Gmail = function(localJQuery) {
 
 
     api.get.storage_info = function() {
-        var div = $(".md.mj").find("div")[0];
-        var used = $(div).find("span")[0].text;
-        var total = $(div).find("span")[1].text;
+        var div = document.querySelectorAll(".md.mj")[0].querySelectorAll("div")[0];
```
josteink, line 361: Iirc querySelector() will select first node only, and is effectively the same as querySelectorAll()[0].

josteink:
Thanks!

Looks good to me, but I will have to test these changes before I can merge. 

I'll try to get it done mondayish 🙂

josteink:
I just tested now, and it seems to have some issues with decimals:

image

Can you look into that? :)

ibercovich:
I couldn't find a a compact way of dealing with commas vs periods for decimal places consistently without writing some more involved code or making a few assumptions. The assumption in this case is that both numbers will either use commas or periods as the decimal separator, but never as the thousands separator, meaning that you will never find 1,000,000.50 MB or 1.000.000,50 MB because it's more clear to simply increase the order of magnitude of the unit to GB. So I can simply replace the commas with periods and leave the rest of the code as it was.

var div = document.querySelector(".md.mj div");
var used = div.querySelectorAll("span")[0].textContent.replace(/,/g, '.'); //convert to standard decimal
var total = div.querySelectorAll("span")[1].textContent.replace(/,/g, '.');
var percent = parseFloat(used.replace(/[^0-9\.]/g, "")) * 100 / parseFloat(total.replace(/[^0-9\.]/g, ""));

josteink:
Yeah. That'll probably work. 

I'd be happy with that.

josteink:
That works for me now! Consider it merged!

josteink:
If you also fix provide a issue and PR with a fix for the message-id issue you encountered, I think that's worthy of a new release :)

PR #639, opened 2021-02-16
I want to be able to get the trimmed part of the body when editing - …

ibercovich:
…using compose.dom('trimmed').val()

If there is a better way, please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1137,7 +1137,7 @@ Compose methods:
 - **.close()** - close compose window
 - **.dom(lookup)** - retrieve preconfigured dom elements for this compose window.
     Lookup can be one of `'to' | 'cc' | 'bcc' | 'id' | 'draft' | 'subject' | 'subjectbox'
-  | 'all_subjects' | 'body' | 'reply' | 'forward' | 'from' | 'send_button'`
+  | 'all_subjects' | 'body' | 'trimmed' | 'reply' | 'forward' | 'from' | 'send_button'`
 
 ### gmail.dom.email(email_el or email_id)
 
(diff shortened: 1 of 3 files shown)


josteink:
I don't see the use for this myself, but I don't see any reason not to add it either.

That said, if you are adding new methods, please also add them to the documentation in `README.md`.

josteink:
Forgot one final part: It also needs to be added to the typescript type-annotations in gmail.d.ts:

https://github.com/KartikTalwar/gmail.js/blob/4f0b48309fa0d8f3c1fee3d3fffa7bd89df877f0/src/gmail.d.ts#L486-L488

And a final question (because I'm obviously going to test this before I merge), what is the expected behaviour of this selector? What part of the compose dialogue is it supposed to extract?

Currently, with this email:
image

I'm getting these results:
image

Is that the expected value?

ibercovich:
Hopefully this screenshot helps clarifying. There might be a better way to accomplish the same thing with existing functionality.

image

josteink:
Yeah that seems to work. Thanks for clearing it up.

Before putting any more work into this, I was wondering if perhaps we could come up with a better name. In my opinion `trimmed`  is not very descriptive, and I certainly didn't understand what to look for.

What about `quoted_reply` instead?

If that's OK, feel free to update the code, the readme and the typescript definitions, and I think we can call it a day :smile:

josteink (approved):

josteink:
Thanks. Merged!

cancan101:
Is there anyway to modify the `quoted_reply` in a way that is respected when the compose is sent?

Ie it looks like calling:
```ts
compose.dom('quoted_reply').val('FOOOO')
```
Does not in fact change the contents of the sent email message.

PR #641, opened 2021-02-17
api.helper.get.is_locale checks only for capitalized 2 letter strings

ibercovich:
Just to clarify from the other thread, I'm assuming the intent was to check only for capitalized 2 letter strings. 

Previous behavior:

```JS
gmail.helper.get.is_locale("US");
true

gmail.helper.get.is_locale("Us");
false

gmail.helper.get.is_locale("uS");
false

gmail.helper.get.is_locale("us");
**true**
```

New behavior:

```JS
gmail.helper.get.is_locale("US");
true

gmail.helper.get.is_locale("Us");
false

gmail.helper.get.is_locale("uS");
false

gmail.helper.get.is_locale("us");
**false**
```

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -174,8 +174,8 @@ var Gmail = function(localJQuery) {
         }
 
         var localePrefix = locale.slice(0, 2);
-        return localePrefix.toLowerCase() === localePrefix ||
-            localePrefix.toUpperCase() === localePrefix;
+        return localePrefix.toUpperCase() === localePrefix;
+            
     };
 
     api.helper.filter_locale = function(locale) {


josteink:

To be clear, I'd like to the match to be 100% case-insensitive. That is, a check for "US", "us", "Us" or "uS" should all succeed.

Gmail.js has a way of being somewhat "fuzzy" about matching and just trying to make things work, and this function should ideally be no different :smiley:

josteink:
Also that bad CI tells me we should be migrating to GitHub Actions too. Yay.

ibercovich:
In that case, shouldn't it just return true once we know the string doesn't contain any numbers? Or do you want some other formal checks?

```js
   api.helper.get.is_locale = function(locale) {
        // A locale is a string that begins with 2 letters, either lowercase or uppercase
        // The "lowercase" check distinguishes locales from other 2-letter strings like "US"
        // (the user"s location?).
        if (!locale || ((typeof locale) !== "string") || locale.length < 2) {
            return false;
        }

        if (locale.match(/[0-9]/)) {
            return false;
        }

        var localePrefix = locale.slice(0, 2);
        // perform additional check ?
        // return localePrefix.toUpperCase() === localePrefix;
        return true;
       
            
    };
```

josteink:
Yeah you're right. Looking at the code now, I must admit I don't remember the exact reason for it being the way it is, and whatever particular quirks I encountered when writing it (basically as a support function for `gmail.get.localization()`).

For now though, I think it's best to leave it as is, unless anyone can pinpoint particular issues the current implementation this is causing.

Sorry for wasting your time :disappointed:

PR #643, opened 2021-02-19
changed to / cc/ bcc selectors to 'input[name=bcc]' instead of textarea

ibercovich:
The current selectors catch the textarea where the user types new email addresses. The existing and fully typed email addresses belong to the input element with the same name.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -3896,9 +3896,9 @@ var Gmail = function(localJQuery) {
         dom: function(lookup) {
             if (!lookup) return this.$el;
             var config = {
-                to:"textarea[name=to]",
-                cc:"textarea[name=cc]",
-                bcc:"textarea[name=bcc]",
+                to:"input[name=to]",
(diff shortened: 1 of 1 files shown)


josteink:
I'm not sure this change is *safe* to be honest.

`input` seems to target the underlying data-layer of the dialog (which at some points may be correct), while `textarea` seems to target the area visible to the user.

Please note these selectors are used for both retrieving and setting data, so while this may work to obtain info from an compose-instance, it may not produce the expected UI-results when using it for *setting* values.

See the following code where these selectors are used:

https://github.com/KartikTalwar/gmail.js/blob/b610092b96122f2efaad65884cd08fe9117b68d2/src/gmail.js#L3790-L3827

josteink:
Consider the following use-case. I've crafted this email:

image

Using the `input` selector will obtain the right address, but it won't replace the existing email when setting:

image

After this, the email still looks the same:

image

Issue #596, opened 2019-11-07
save_draft message not working with new data layer


morungos:
I've noticed that the save_draft message hasn't been working since Gmail 2018 came on the scene. We needed it, so I've done a little digging and found a simple fix that seems to work. 

The patch is, approximately:

    @@ -1437,6 +1437,8 @@ var Gmail = function(localJQuery) {
                         events.send_message = [params.url, params.body, sent_email];
                     } else if (prop.indexOf("^scheduled") > -1) {
                         events.send_scheduled_message = [params.url, params.body, sent_email];
    +                } else if (prop.indexOf("^r") !== -1 && prop.indexOf("^r_bt") !== -1) {
    +                    events.save_draft = [params.url, params.body, sent_email];
                     }
                 }
             }

I'm happy to make this into a PR if it's useful, but any advice on writing tests for it would be very welcome. I did this by tracking a bunch of the POST requests to /i/s endpoints, which seems to be where the action is. 

I have no guarantees that I've caught everything, but it's looking like it does the right thing with this fix in place. And I have managed to use this to do what I needed, which is to cruft out some of the automatically injected tags that I was adding in the UI but didn't want Gmail to write into its storage systems. 

Feedback, thoughts, and comments welcome.

josteink:
> I've noticed that the save_draft message hasn't been working since Gmail 2018 came on the scene.

That a shame, but no surprise. Nobody has actively worked on fixing *all* the observables since new Gmail came out.

> I'm happy to make this into a PR if it's useful

Great!

> any advice on writing tests for it would be very welcome. I did this by tracking a bunch of the POST requests to /i/s endpoints, which seems to be where the action is.

Basically our test-cases so far is not so much about the request interception, but at the parsing-stage once we do have the relevant JSON or protobuffer-data.

We typically do this by putting a JSON sample in `testdata-parser.js` and add the test-case itself to `test.parsing.js`.

You can see some examples for the kind of tests we have here:

https://github.com/KartikTalwar/gmail.js/blob/c7ce9aa7d0335333cde4846d2c65057bb7df16ad/test/test.parsing.js#L469-L483

It's pretty basic stuff, but it's better than nothing.

PR #653, opened 2021-09-26
Persist if a message is a draft during request parsing

cbonnoit:
Hi - I found it useful to be able to check if a message in `gmail.cache.emailIdCache` was draft or not. The below change seems to accomplish that for me (tested in Gmail with new messages and replies to existing conversations, as both a free and Gmail Business user).

I do not have any insight into what these labels are or how stable they will be, but FWIW they were also mentioned in a similar context in https://github.com/KartikTalwar/gmail.js/issues/596.

Hope this is useful and please advise if you are aware of a better approach to adding this information - happy to revise if so.

Thanks


josteink:
Hey there and thanks for the PR!

The change itself looks simple and safe enough. 

I have some (very) few comments on the implementation, but in essence it looks good. 

Before approving though, I'll have to test the changes myself to make sure things works as expected. 

I should hopefully get that done sometime next week 🙂

cbonnoit:
Great, thanks for taking a look

josteink:
Great feature! I've just tested it and it works like a charm! 

Which leaves us with the actual code-review before merging :smile:

src/gmail.js
```diff
@@ -1205,6 +1205,11 @@ var Gmail = function(localJQuery) {
         return res;
     };
 
+    api.tools.parse_fd_bv_is_draft = function(item) {
+        if (!Array.isArray(item)) return null;
```
josteink, line 1209: We want our `is_draft`-check to return a `boolean`. it should be either `true` or `false`.

Don't return `null` here when you can return `false` instead.

src/gmail.js
```diff
@@ -1329,6 +1334,8 @@ var Gmail = function(localJQuery) {
                     const fd_email_timestamp = Number.parseInt(fd_email["2"]["17"]);
                     const fd_email_date = new Date(fd_email_timestamp);
 
+                    const fd_email_is_draft = api.tools.parse_fd_bv_is_draft(fd_email2["4"]);
```
josteink, line 1337: In this part of the codebase, in this particular function, you have to keep in mind that we are actually intercepting/hijacking Gmail XHR requests, and any failure in our code at this point will cause entire Gmail to stop working. Full stop.

**As such, this code needs to be 100% bullet-proof against any sort of failure.**

We can kinda accept that Gmail.JS will stop working, but we should never accept that using Gmail.JS in an extension causes *Gmail itself* to stop working. At least, that has been the priority for this part of the code so far.

So, with that said: This new line seems to be one of few using the `fd_email2` variable, the other being `api.tools.parse_fd_bv_contact()`.

I'd be more comfortable with adding this new function into this loop, if it implements the same kind of safeguards as `api.tools.parse_fd_bv_contact()` does (that is, `try{ ... } catch(){ return false; }`) .

josteink:
This PR expands the available properties available on "new" email-data. As such the type-definitions for typescript users should be updated too:

https://github.com/KartikTalwar/gmail.js/blob/fce0ae72e3aaf23cfa96ce1e02a1d9a35be39706/src/gmail.d.ts#L910-L926

josteink:
This PR implements a new parser for raw Gmail XHR data.

Would you be able to add a test-case for it, among the other test-cases we have?

https://github.com/KartikTalwar/gmail.js/tree/master/test/testdata-parser-json

If you could add 2 test-data instances (one draft, one not) and write tests to verify that the Gmail.js parser actually handles those correctly, your new functionality will be much better protected against regressions in the future too :smiley:

josteink:
And that's it!

If you can take a look at those things, I'll be *more* than happy to merge these changes :)

src/gmail.js
cbonnoit: Thanks - totally agree on the philosophy.  I did not think it was possible for this code to cause a failure - I think we're guaranteed `fd_email2["4"]` itself will not throw because (a) we already called `fd_email2["2"]` for `parse_fd_bv_contact` and (b) everything extends an object so getting a property is always safe. If it is not of type `Array`, then `Array.isArray` will return `false` and we'll exit.  If it is of type `Array` then `.includes` will be well defined. I mean, I suppose it's possible someone overrides `Array` to remove `include`, but that seemed fairly intentionally malicious at that point :)

Anyway, let me know if you think I've missed something here - also happy to just add it anyways if you feel strongly since its presumably irrelevant from a performance standpoint.

src/gmail.js
cbonnoit: The reason I returned `null` is that the item not being an array means the schema has changed (a field that used to always be array-like is no longer populated with array data).  In this case, we really don't know if the message as a draft here, so it seemed nicer to say `null` (not calculable / know) rather than `false` (known to be not a draft). Anyway, if you find this argument unconvincing I don't mind changing!

cbonnoit:
Thanks - couple comments on specific issues above.  Good catch on the `.ts` and I'll look into test cases.

josteink:
I’ve also enabled CI for this PR. It seems to have issue with a undefined variable. Guess you should look into that too 🙂

josteink:
If indeed git master is broken, that obviously needs to be fixed. 

I’ll let you know when I have that done and hopefully you can merge that to make your build green too.

cbonnoit:
Sure, sounds good.

josteink:
Error in git master should be fixed now. And that fixes a bug too :)

cbonnoit:
Hi @josteink - I've updated the `.ts` and added a couple test cases. I think we still have two open threads above (`null` values and a `try` block) - I didn't change here per my comments, but let me know if you feel strongly and I don't mind switching.  Other than that, I think this review should be good to go.

Thanks

src/gmail.js
josteink: Nobody is going to check for null. Making it false is more consistent with how we otherwise do things. 

Right now the typescript definitions also say this field is Boolean only, so let’s make it so 😉

src/gmail.js
josteink: In theory I agree with you. I’m practice, let’s add a try/catch just in case 😅

josteink:
Looks great! Let’s fix those two nitpicks and I think we have this landed!

cbonnoit:
Hi @josteink , this is updated per your requests.

josteink:
Thanks for the PR! Merged and published!

cbonnoit:
Great, glad it was useful & thanks for working with me on it

PR #655, opened 2021-10-26
fix: access To, Cc, Bcc elements in new Gmail contact compose

### LINKED ISSUE #654: Lost the ability to get to, cc and bcc fields from email compose
It looks like gmail did an update, and changed the compose window.
composeEl.to() will not give an error
Screenshot 2021-10-22 at 12 52 07

This also means that gmail.observe.on('recipient_change' no longer triggers

Any workaround for this, or fix in the works?

huksley:
Fixes #654 

https://workspaceupdates.googleblog.com/2021/10/visual-updates-for-composing-email-in-gmail.html

* **Rapid Release Release domains**: Extended rollout (potentially longer than 15 days for feature visibility) starting on October 20, 2021. 
* **Scheduled Release domains**: Extended rollout (potentially longer than 15 days for feature visibility) beginning no earlier than November 8, 2021.

So old access need to be supported indefinitely? :(

### DIFF
--- a/src/gmail.d.ts
+++ b/src/gmail.d.ts
@@ -739,7 +739,7 @@ declare type GmailBindAction =
     | 'add_to_tasks' | 'move_label' | 'save_draft' | 'discard_draft' | 'send_message' | 'send_scheduled_message'
     | 'expand_categories' | 'delete_label' | 'show_newly_arrived_message' | 'poll'
     | 'new_email' | 'refresh' | 'open_email' | 'upload_attachment' | 'compose'
-    | 'compose_cancelled' | 'recipient_change' | 'recipient_change_new' | 'view_thread' | 'view_email'
+    | 'compose_cancelled' | 'recipient_change' | 'recipient_change' | 'view_thread' | 'view_email'
     | 'load_email_menu';
 
 interface GmailObserve {
--- a/src/gmail.d.ts
+++ b/src/gmail.d.ts
@@ -739,7 +739,7 @@ declare type GmailBindAction =
     | 'add_to_tasks' | 'move_label' | 'save_draft' | 'discard_draft' | 'send_message' | 'send_scheduled_message'
     | 'expand_categories' | 'delete_label' | 'show_newly_arrived_message' | 'poll'
     | 'new_email' | 'refresh' | 'open_email' | 'upload_attachment' | 'compose'
-    | 'compose_cancelled' | 'recipient_change' | 'recipient_change' | 'view_thread' | 'view_email'
+    | 'compose_cancelled' | 'recipient_change' | 'view_thread' | 'view_email'
     | 'load_email_menu';
 
 interface GmailObserve {
--- a/src/gmail.d.ts
+++ b/src/gmail.d.ts
@@ -739,7 +739,7 @@ declare type GmailBindAction =
     | 'add_to_tasks' | 'move_label' | 'save_draft' | 'discard_draft' | 'send_message' | 'send_scheduled_message'
     | 'expand_categories' | 'delete_label' | 'show_newly_arrived_message' | 'poll'
     | 'new_email' | 'refresh' | 'open_email' | 'upload_attachment' | 'compose'
-    | 'compose_cancelled' | 'recipient_change' | 'view_thread' | 'view_email'
+    | 'compose_cancelled' | 'recipient_change' | 'recipient_change_new' | 'view_thread' | 'view_email'
     | 'load_email_menu';
 
 interface GmailObserve {
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -4000,7 +4000,6 @@ var Gmail = function(localJQuery) {
                     }
                 })
 
-                console.info("elems", elems)
                 const el = elems[lookup]
                 if (!el) api.tools.error("Dom lookup failed. Unable to find \"" + lookup + "\"", config, lookup);
                 return el;
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -2374,16 +2374,19 @@ var Gmail = function(localJQuery) {
                         // console.log("recipient timeout handler", api.tracker.recipient_matches.length);
                         if(!api.tracker.recipient_matches.length) return;
 
-                        // determine an array of all emails specified for To, CC and BCC and extract addresses into an object for the callback
-                        let composeRoot = api.tracker.recipient_matches[0].closest("div.M9");
-                        // sometimes (on copy-paste of contact in peoplekit mode) element disappears
-                        if (composeRoot.length === 0 && api.tracker.recipient_matches.length > 0) {
-                            composeRoot = api.tracker.recipient_matches[1].closest("div.M9");
-                        }
+                        let composeRoot = [];
+                        // sometimes (on copy-paste of contact in peoplekit mode) element disappears so iterate for all matches
+                        api.tracker.recipient_matches.forEach(match => {
+                            if (composeRoot.length === 0) {
+                                composeRoot = match.closest("div.M9");
+                            }
+                        });
+
                         if (composeRoot.length === 0) {
-                            api.tools.error("Can't find composeRoot for " + match)
+                            api.tools.error("Can't find composeRoot for " + match);
                         }
                         var compose = new api.dom.compose(composeRoot);
+                        // determine an array of all emails specified for To, CC and BCC and extract addresses into an object for the callback
                         var recipients = compose.recipients();
                         callback(compose, recipients, api.tracker.recipient_matches);
 
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -2314,7 +2314,7 @@ var Gmail = function(localJQuery) {
     // console.log( "Observer set for", action, callback);
     api.observe.initialize_dom_observers = function() {
         api.tracker.dom_observer_init = true;
-        api.tracker.supported_observers = ["view_thread", "view_email", "load_email_menu", "recipient_change", "compose"];
+        api.tracker.supported_observers = ["view_thread", "view_email", "load_email_menu", "recipient_change", "recipient_change_new", "compose"];
         api.tracker.dom_observers = {
 
             // when a thread is clicked on in a mailbox for viewing - note: this should fire at a similar time (directly after) as the open_email XHR observer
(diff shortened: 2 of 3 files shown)


src/gmail.js
```diff
@@ -2314,7 +2314,7 @@ var Gmail = function(localJQuery) {
     // console.log( "Observer set for", action, callback);
     api.observe.initialize_dom_observers = function() {
         api.tracker.dom_observer_init = true;
-        api.tracker.supported_observers = ["view_thread", "view_email", "load_email_menu", "recipient_change", "compose"];
+        api.tracker.supported_observers = ["view_thread", "view_email", "load_email_menu", "recipient_change", "recipient_change_new", "compose"];
```
josteink, line 2317: I don't think we wan't to add another event here. The best would be if existing code kept working unmodified by simply updateing `gmail.js`.

Do you think that could be possible?

src/gmail.js
```diff
@@ -2532,9 +2561,15 @@ var Gmail = function(localJQuery) {
...
+                                let observer = api.tracker.dom_observer_map["afV"];
+                                let handler = api.tracker.dom_observers.recipient_change_new.handler;
+                                api.observe.trigger_dom(observer, $(mutation.target), handler);
+                            }
+
                             if (removedNode.className === "vR") {
-                                var observer = api.tracker.dom_observer_map["vR"];
-                                var handler = api.tracker.dom_observers.recipient_change.handler;
+                                let observer = api.tracker.dom_observer_map["vR"];
+                                let handler = api.tracker.dom_observers.recipient_change.handler;
```
josteink, line 2572: Good catch!

src/gmail.js
```diff
@@ -3984,6 +4036,19 @@ var Gmail = function(localJQuery) {
...
                 show_bcc: "span.aB.gQ.pB"
             };
+
+            // Post "peoplekit"
+            if (this.$el.find("textarea[name=to]").length === 0 && ["to", "cc", "bcc"].includes(lookup)) {
+                const configPeoplekit = {
+                    to:"div[name=to] input",
+                    cc:"div[name=cc] input",
+                    bcc:"div[name=bcc] input"
+                };
```
josteink, line 4046: I think it would be cleaner if we instead of having our own "full" branch for the entire code-sequence (lookup + retrieve), just used it to patch the `config` variable we already have.

That way we have 1 map, 1 patch and 1 general code-path for all lookups.

src/gmail.js
```diff
@@ -3984,6 +4036,19 @@ var Gmail = function(localJQuery) {
                 show_cc: "span.aB.gQ.pE",
                 show_bcc: "span.aB.gQ.pB"
             };
+
+            // Post "peoplekit"
+            if (this.$el.find("textarea[name=to]").length === 0 && ["to", "cc", "bcc"].includes(lookup)) {
```
josteink, line 4041: Would be possible to extract the "peoplekit" check into another method, and *if* we are doing people-kit we patch up the config, no matter what the current lookup is?

src/gmail.js
```diff
@@ -3843,19 +3878,36 @@ var Gmail = function(localJQuery) {
...
-                if(options.flat) {
-                    recipients.push(recipient.value);
-                } else {
-                    if(!recipients[recipient.name]) recipients[recipient.name] = [];
-                    recipients[recipient.name].push(recipient.value);
-                }
-            });
-            return recipients;
+            // Peoplekit
+            if (this.$el.find("div[data-hovercard-id]").length) {
```
josteink, line 3883: Again: Could we extract these checks into a more general function/method which can be reused elsewhere, or is it so highly contextual that we cannot detect this without a specific UI-setting currently active?

src/gmail.js
```diff
@@ -3843,19 +3878,36 @@ var Gmail = function(localJQuery) {
...
+                var recipients = options.flat ? [] : { to: [], cc: [], bcc: [] };
+                this.$el.find(".GS input[type=hidden]"+name_selector).each(function(idx, recipient ){
+                    if(options.flat) {
+                        recipients.push(recipient.value);
+                    } else {
+                        if(!recipients[recipient.name]) recipients[recipient.name] = [];
+                        recipients[recipient.name].push(recipient.value);
+                    }
+                });
+                return recipients;
```
josteink, line 3909: There's quite a bit of duplication here, of what I will admit I find to be non-great code, and doubly admit it's definitely not your fault.

What bothers me most right now is probably the flat vs non-flat aspect, and that seems like something which could be split out quite easily:

````javascript
let recipients = [];

if (isPeopleKit()) {
    recpieints = this.$el.find("peopleKit-selector).map( i => i.getAttribute(""));
} else {
    recpieints = this.$el.find("nonpeopleKit-selector);
}

if (options.flat) {
    return recipients;
} else {
    let result = {};
    result[options.type] = recipients;
    return result;
}
````

I think that would be much cleaner. What do you think?

File: src/gmail.js [hunk 7]
```diff
@@ -3984,6 +4036,19 @@ var Gmail = function(localJQuery) {
                 show_cc: "span.aB.gQ.pE",
                 show_bcc: "span.aB.gQ.pB"
             };
+
+            // Post "peoplekit"
+            if (this.$el.find("textarea[name=to]").length === 0 && ["to", "cc", "bcc"].includes(lookup)) {
+                const configPeoplekit = {
+                    to:"div[name=to] input",
```
josteink, line 4043: Not having seen the UI for the "peoplekit" update... Are these `input`-element hidden? Can the selector be used for updating values or just extracting them?

josteink:
Sorry for the ultra-slow feedback, but now I've taken a basic look-over and think we have somethink to work on.

Let me know if you have any questions :)

huksley:
Fixed all the comments also dropped "recipient_change_new" stuff - now it is in one single event for both new and old Gmail contact mode. Thankfully I have access now to both versions.

src/gmail.d.ts
```diff
@@ -739,7 +739,7 @@ declare type GmailBindAction =
     | 'add_to_tasks' | 'move_label' | 'save_draft' | 'discard_draft' | 'send_message' | 'send_scheduled_message'
     | 'expand_categories' | 'delete_label' | 'show_newly_arrived_message' | 'poll'
     | 'new_email' | 'refresh' | 'open_email' | 'upload_attachment' | 'compose'
-    | 'compose_cancelled' | 'recipient_change' | 'recipient_change_new' | 'view_thread' | 'view_email'
+    | 'compose_cancelled' | 'recipient_change' | 'recipient_change' | 'view_thread' | 'view_email'
```
josteink, line 742: This change is not needed, as that event-type is already defined :)

src/gmail.js
```diff
@@ -2368,7 +2375,15 @@ var Gmail = function(localJQuery) {
...
 
                         // determine an array of all emails specified for To, CC and BCC and extract addresses into an object for the callback
-                        var compose = new api.dom.compose(api.tracker.recipient_matches[0].closest("div.M9"));
+                        let composeRoot = api.tracker.recipient_matches[0].closest("div.M9");
+                        // sometimes (on copy-paste of contact in peoplekit mode) element disappears
+                        if (composeRoot.length === 0 && api.tracker.recipient_matches.length > 0) {
+                            composeRoot = api.tracker.recipient_matches[1].closest("div.M9");
+                        }
+                        if (composeRoot.length === 0) {
+                            api.tools.error("Can't find composeRoot for " + match)
```
josteink: Keep in mind this code is running in the background while intercepting genuine Gmail API traffic. We do NOT want the presence of Gmail.JS to crash Gmail under *any* circumstances.

Having the Gmail.JS API fail is OK. Having the extension depending on Gmail.JS fail is OK. Having Gmail itself fail is *never* OK.

As such, I'd rather have this statement replaced by a generic `console.error()` statement, rather than calling into something which may or may not throw :smile:

src/gmail.js
```diff
@@ -2532,9 +2547,15 @@ var Gmail = function(localJQuery) {
                         var removedNodes = mutation.removedNodes;
                         for (var j = 0; j < removedNodes.length; j++) {
                             var removedNode = removedNodes[j];
+                            if (removedNode.className === "agh" && removedNode.querySelector("div[data-hovercard-id]")) { // contains recipient in peoplekit
+                                let observer = api.tracker.dom_observer_map["afV"];
+                                let handler = api.tracker.dom_observers.recipient_change.handler;
+                                api.observe.trigger_dom(observer, $(mutation.target), handler);
+                            }
+
                             if (removedNode.className === "vR") {
```
josteink, line 2559: I'm not sure how that could happe, but just to prevent the same event from triggering twice, would it make sense to make this an `else if` instead?

It would also make it clearer in code that those 2 events/code-paths should never occur at the same time.

src/gmail.js
```diff
@@ -3843,23 +3864,45 @@ var Gmail = function(localJQuery) {
...
-            var recipients = options.flat ? [] : { to: [], cc: [], bcc: [] };
-            this.$el.find(".GS input[type=hidden]"+name_selector).each(function(idx, recipient ){
-                if(options.flat) {
-                    recipients.push(recipient.value);
+            const peopleKit = api.check.is_peoplekit_compose(this.$el);
+            
+            const type_selector = options.type ?
+                peopleKit ?
+                    "div[name=" + options.type + "] " :
+                    "[name=" + options.type + "]" : "";
```
josteink, line 3875: Right now we can provide a more specific selector for peoplekit (`div` element), but is it actually *required*?

If we don't need it, the code without that variance would be a lot simpler.

src/gmail.js
```diff
@@ -3843,23 +3864,45 @@ var Gmail = function(localJQuery) {
...
+                    ["to", "cc", "bcc"].forEach(type => {
+                        result[type] = found.toArray()
+                            .filter(r => r.type === type)
+                            .map(r => r.email);
+                    });
                 }
-            });
-            return recipients;
+                return result;
+            }
```
josteink, line 3903: This is all in all a lot better. I like that :)

src/gmail.js
```diff
@@ -3843,23 +3864,45 @@ var Gmail = function(localJQuery) {
...
         },
 
         /**
-           Retrieve the current "to" recipients
+           Retrieve the typing area for "to" recipients, not recipients. 
+           Either textarea or input, which can be empty if last recipient are typed and selected (by pressing ENTER)
```
josteink, line 3908: The documentation is very clear on "retrieve", but the API code in the past has been bi-directional.

That is, these calls could get *and* set the value in the approiate compose-field. Is this still true with this PeopleKit implementation?

josteink (changes requested):
Good improvement on last run.

Still have a few things to iron out, but getting closer :smile:

src/gmail.js
huksley: It is not even true with pre-peoplekit - textarea this thing refers to gets cleared immediately once you enter something and press ENTER

So you can obtain only intermediate value, but you can set value

src/gmail.js
huksley: This was improved, please check. It should not affect GMail as it just does nothing if composeRoot not found

src/gmail.js
josteink: Oh well. In that case it still sounds like an improvement, and I'm not going to block this PR over this :smile:

src/gmail.js
josteink: The new code does indeed look better. Thanks!

src/gmail.js
josteink: Looking at this again, this seems not to be related to XHR interception at all (my bad), and in that case this code is perfectly acceptable. And the updated code is even better.

I'm OK with this as things are now :+1:

src/gmail.js
```diff
@@ -295,6 +295,13 @@ var Gmail = function(localJQuery) {
         return check_1.length > 1 || check_2.length > 1;
     };
 
+    /**
+    * New contact selection UI as announced in 
+    * https://workspaceupdates.googleblog.com/2021/10/visual-updates-for-composing-email-in-gmail.html
+    **/ 
+    api.check.is_peoplekit_compose = function (el) {
+        return !!$(el).find("div[name=to] input[peoplekit-id]").length;
```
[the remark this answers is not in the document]
josteink: Looking better now. Thanks!

src/gmail.js
josteink: Great!

josteink (approved):
Thanks for hanging in there. Consider it merged!

josteink:
Fix should be available in new release `gmail-js@1.0.15`. Thanks @huksley !

PR #658, opened 2021-10-29
feat: get compose attachments

### LINKED ISSUE #657: Access Compose Attachments
PR coming soon...

ek23:
Closes #657

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -366,7 +366,7 @@ var Gmail = function(localJQuery) {
     };
 
 
-    api.get.storage_info = function() {        
+    api.get.storage_info = function() {
         var div = document.querySelector(".md.mj div");
         var used = div.querySelectorAll("span")[0].textContent.replace(/,/g, '.'); //convert to standard decimal
         var total = div.querySelectorAll("span")[1].textContent.replace(/,/g, '.');
@@ -1213,7 +1213,7 @@ var Gmail = function(localJQuery) {
         catch (e) {
             return false;  // warning: case not seen during testing and value is untrustworthy
         }
-        
+
     };
 
     api.tools.parse_fd_bv_contact = function(item) {
@@ -1309,7 +1309,7 @@ var Gmail = function(localJQuery) {
             return {};
         }
     };
-    
+
     api.tools.parse_fd_request_payload = function(json) {
         // ensure JSON-format is known and understood?
         let thread_root = json["2"];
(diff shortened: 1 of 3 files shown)


src/gmail.js
```diff
@@ -3932,6 +3932,33 @@ var Gmail = function(localJQuery) {
...
+            var out = [];
+            var failed = false;
+
+            this.dom("attachments").each(function() {
+                var el = $(this);
+
+                var result = {};
+                result.$el = el;
+                result.name = el.find("div.vI").html();
+                result.size = el.find("div.vJ").html();
```
josteink, line 3948: From the outside as an API consumer, this property is easy to assume to be numerical, but looking at the code, it looks like its a displayvalue ("5mb", etc).

Could we name it better to avoid incorrect usage?

josteink:

This looks like great stuff, and I'm all for it. You cleaning up all whitespaces and stuff creates  bit of a messy PR, but I guess it's worth just getting done, so don't sweat it over that.

Please note that it seems like you have extended the API with new calls and new types, and when we do that, we also need to update [the type-definitions](https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.d.ts). Can you handle that? :smile: 

Could you bring them in line with your code-changes, and maybe also add your new functional call to [the readme](https://github.com/KartikTalwar/gmail.js/blob/master/README.md#gmaildomcomposecompose_el)?

src/gmail.js
ek23: I think this is ok.  The return type is GmailDomAttachment where size is a string.  This matches what gmail.dom.email.attachments() returns

ek23:
Updated the type definitions and readme

josteink (approved):

josteink:
Looks good to me. Let’s rock! Thanks for contributing!

src/gmail.js
jotdhaliwal: @ek23 @josteink I am having google chrome extension , in which I want to get the attachment of the email and want to download those into my system. is this possible to get URL of the attachment via your API

josteink: Yes. Check api.new.email_data(). It has information about extensions, including URL.

jotdhaliwal: @josteink can you please give me a example code as I tried to ejected your gmail.js into my DOM but it returns nothing , little guidance will be appreciated

josteink: Use the supplied boilerplate project. 

Load Gmail and a specific email thread. Go to the web CI sile and type:

    var attachements = gmail.new.get.email_data().attachments;

PR #664, opened 2022-02-03
fix: disable checking for SID, ik, act for http_event, typescript types

huksley:
Currently,  gmail.js checks for those deprecated fields (probably from old Gmail API)
as a result, there is only 2 kinds of requests which can be received by calling on("http_event") or after("http_event"): `https://mail.google.com/mail/u/0/?ui=2&ik=` and `https://mail.google.com/mail/u/0/flushimpressions?ui=2&ik=` and does not include `https://mail.google.com/mail/u/0/i/fd`- fetch mail data requests.

This PR fixes that so all other kinds of XHR requests can be accessed from gmail.js without the need to implement XHR hooks myself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 
 # Changelog
 
+## Version 1.0.16
+
+- Fix `gmail.observe.on("http_event")` and `gmail.observe.after("http_event")` triggers to receive all XHR requests.
+
 ## Version 1.0.15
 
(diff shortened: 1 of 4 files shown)


josteink:

Change looks good to me. You might want to add it to the changelog too? :)

huksley:
@josteink Added to CHANGELOG as v.1.0.16

josteink:
Looks good. Consider it merged!

Issue #662, opened 2022-01-27
api.tools.parse_requests mutating the xhrParams


kinkoazc:
Hello,

I came across an issue involving the use of `gmail.js` library by 2(or more) currently installed extensions.
I think that they come into conflict around the mutable changes done to the `params` variable, in the `api.tools.parse_requests` method.

Scenario:
1. chrome extension no. 1 patches the `send` method, `api.tools.patch(win.XMLHttpRequest.prototype.send`;
2. chrome extension no. 2 also patches the `send` method(now, already patched by chrome extension no. 1), `api.tools.patch(win.XMLHttpRequest.prototype.send`;
3. a XHR request gets made, with `this.xhrParams` initially as `{method: 'POST', url: 'https://mail.google.com/sync/u/0/i/s?hl=en&c=21'}`;
4. chrome extension no. 2's patch comes into play, executes `api.tools.parse_requests`; now `this.xhrParams` looks like `{method: 'POST', url_raw: 'https://mail.google.com/sync/u/0/i/s?hl=en&c=21', body_params: {...}, url: {hl: 'en', c: '21'}, body: '...'}`;
5. chrome extension no. 2's patch sends the request, `var out = orig.apply(this, arguments);`;
6. now chrome extension no. 1 patch comes into play, receiving, as `this.xhrParams`, the above `{method: 'POST', url_raw: 'https://mail.google.com/sync/u/0/i/s?hl=en&c=21', body_params: {...}, url: {hl: 'en', c: '21'}, body: '...'}`;
7. in chrome extension no. 1's `api.tools.parse_requests` method `this.xhrParams` will come out as `{method: 'POST', url_raw: {hl: 'en', c: '21'}, body_params: {...}, url: {}, body: '...'}`, which will mess with the subsequent library logic, e.g. what happens in `api.tools.parse_request_payload`.

Is this behavior correct/known? Thank you.

josteink:
Oh bugger. 

I really did my best to rewrite Gmail.js to be “stackable” without causing issues with concurrent extensions using it. 

Guess there’s more work which needs to be done then.

Not sure when I will have time to go in depth into this. If you can find a way to avoid this causing issues, I’m open to advice. 

And obviously: PRs are always welcome 🙂

josteink:
Hey there and thanks for the detailed report. 

This issue was not really known, to be honest. 

In other words: Oh bugger. 

[...]

PR #666, opened 2022-02-06
fix: api.tools.parse_requests mutating the xhrParams (#662)

kinkoazc:


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 
 # Changelog
 
+## Version 1.0.17
```
josteink, line 4: Version 1.0.16 is not yet released. Feel free to add this to that section.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "gmail-js",
-    "version": "1.0.15",
+    "version": "1.0.17",
```
josteink, line 3: 1.0.16 would fit better here 🙂

package.json
```diff
@@ -9,9 +9,9 @@
         "url": "https://github.com/KartikTalwar/gmail.js.git"
     },
     "scripts": {
-        "lint": "./node_modules/.bin/eslint src/*.js",
-        "tsc": "./node_modules/.bin/tsc",
-        "test": "./node_modules/.bin/mocha test/test.*.js",
+        "lint": "npx eslint src/*.js",
+        "tsc": "npx tsc",
+        "test": "npx mocha test/test.*.js",
```
josteink, line 14: Are these changes strictly speaking required for this PR?

package.json
```diff
@@ -23,6 +23,7 @@
         "gmail firefox extension"
     ],
     "dependencies": {
+        "gmail-js": "^0.9.9",
```
josteink, line 26: Are we… a dependency on ourself? That sounds weird 😃

package.json
kinkoazc: The scripts, as they were, didn't work in my case, so I've replaced them with `npx`, which first looks for local versions of the respective modules. I can revert them if it's not ok for some reason.

src/gmail.js
```diff
@@ -64,7 +64,7 @@ var Gmail = function(localJQuery) {
         console.warn("GmailJS: using deprecated API for old Gmail.", text);
     }
 
-    api.version           = "0.8.0";
+    api.version           = "1.0.17";
```
josteink, line 67: api version != package version. 

Also: this property haven’t really been used (or relevant at all) since support for bundling or stacking was introduced. You can safely omit this change.

package.json
kinkoazc: I've included the old library for testing purposes. Should have been part of the `devDependencies` group I guess.

src/gmail.js
kinkoazc: Will do.

src/gmail.js
```diff
@@ -1995,24 +2002,41 @@ var Gmail = function(localJQuery) {
                 var events = false;
                 if (this.xhrParams) {
                     this.xhrParams.body = body;
+
+                    // restore original values of xhrParams, if they were altered by upstream instances of gmail.js
+                    if (typeof this.xhrParams.url !== 'string') {
+                        if (
+                            this
```
josteink, line 2009: I appreciate the “safety first” approach in this critical code-path, but can “this” ever be null here?

package.json
kinkoazc: Ok.

src/gmail.js
kinkoazc: Yes, you're right. Will remove.

package.json
josteink: If this works better, I’m for that. 

That said, this would make npx a new devDependency we didn’t have before, and it will have to be declared as such too.

package.json
josteink: Yeah devDependency makes much more sense. Nice test too!

package.json
kinkoazc: `npx` comes with `npm`, https://docs.npmjs.com/cli/v7/commands/npx. So no new dependency.

CHANGELOG.md
kinkoazc: Will do.

package.json
josteink: TIL. No problem then!

josteink:
Thanks for taking the time to make this PR. That’s greatly appreciated!

Having reviewed all the superficial stuff I can find (which is easy to do on a phone), I’ll try to review the actual core of your changes later on and try to help get it merged. 

If you don't hear anything in a few days time, don’t be afraid to ping me. Sometimes I forget, and in those case a gentle reminder is absolutely appreciated!

kinkoazc:
Sure, will do.
Thank you as well for your work!

josteink:
Looks great to me. Much better than I would have done it. Thanks for putting in the effort :blush: 

Consider it merged!

Issue #667, opened 2022-02-07
`visible_messages()` breaks application in certain cases


mhatvan:
I think this occurs when there are ads in a Gmail tab (primary, social, etc):

```
gmailJsLoader.js:3 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'dataset')
    at HTMLTableRowElement.<anonymous> (gmailJsLoader.js:3:50302)
    at Function.each (gmailJsLoader.js:3:60064)
    at b.fn.init.each (gmailJsLoader.js:3:58543)
    at Object.o.dom.visible_messages (gmailJsLoader.js:3:50078)
    at checkForVisibleMessages (extension.js:13:42)
    at addCustomDOMContent (extension.js:155:27)
    at extension.js:100:13
    at c (m=b:869:118)
```

josteink:
Hey there and thanks for reporting!

Looking at the stack trace, is this just using the plain boilerplate project, or something more.. “orchestrated”? 🙂

mhatvan:
That's the plain gmail.js code in a file. I was able to fix this locally by adding checks to the source code for that method.

josteink:
Right. In which case, feel free to submit a PR 😄

josteink:
This should now be published in `v1.0.17` on npmjs.com.

PR #668, opened 2022-02-09
fix: `visible_messages()` breaks application in certain cases

mhatvan:
## Description

- check for existence of `idNode` or fallback to empty string in `visible_messages()`

Closes https://github.com/KartikTalwar/gmail.js/issues/667


josteink:
That looks good to me. Consider it merged! Thanks for the fix!

Issue #673, opened 2022-04-04
Attachments from fd_embedded_json are not correctly parsed


onestep:
**Prerequisites:**
- Account with `_GM_setData` preload.

**Steps to reproduce:**
1. open Inbox folder
2. receive email with attachment from some another account
3. reload page
4. navigate to new received email

**Actual result:**
Attachments data are processed incorrectly: attachment_id is `"a"`, other properties are `undefined`.

onestep:
I've opened a pull request #674 with a fix. I decided to keep `parse_fd_attachments` as is, but it can be renamed to `parse_fd_request_payload_attachments` if needed.

josteink:
Fixed in #673 and published on npmjs as v1.0.18.

PR #674, opened 2022-04-04
fix: support parsing attachments data from fd_embedded_json (#673)

onestep:


src/gmail.js
```diff
@@ -1436,8 +1454,7 @@ var Gmail = function(localJQuery) {
                     //TODO : need a refactoring
                     const fd_email_content_html = api.tools.parse_fd_embedded_json_content_html(fd_email);
 
-                    //TODO
-                    const fd_attachments = api.tools.parse_fd_attachments(fd_email["12"]);
+                    const fd_attachments = api.tools.parse_fd_embedded_json_attachments(fd_email["12"]);
```
josteink, line 1457: How compatible is this code? Shouldn’t we have fallback to the old code, if the new code fails?

onestep: I see no reason for this, since previous code was added just as a stub, and was marked with TODO. We have the same code, as I've added here, in production for more than a year and it works fine.
I've also extended tests for all parsing cases to check for parsed attachments, they pass fine as well.

josteink: Ah right. Fair enough!

josteink (approved):

josteink:
Looks good to me. Consider it merged!

PR #680, opened 2022-04-25
Fixed compose buttons being duplicated when using add_compose_button multiple times

DrParanoia:
I you try to add two buttons using `add_compose_button` you will get three, if you try to add three buttons, you will get 6 and so on...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 
 # Changelog
 
+## Version 1.0.19
+
+- Fix compose button being duplicated when using `gmail.tools.add_compose_button` more than once.
+
 ## Version 1.0.18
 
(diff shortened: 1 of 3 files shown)


josteink:

The change looks reasonable to. Thanks for the fix!

Could also bump the version and add your fix to the change log?

DrParanoia:
@josteink Sure, just a sec

DrParanoia:
@josteink Done! 😄

josteink:
Thanks. Great! Consider it merged!

I'll try to get a new version out to npm js later today 🙂

DrParanoia:
@josteink Thanks a bunch!

josteink:
Now published on npmjs as 1.0.19!

https://www.npmjs.com/package/gmail-js

PR #682, opened 2022-04-27
Fix typing on the after send_message signature to include response

cancan101:
See docs here: https://github.com/KartikTalwar/gmail.js#gmailobserveafteraction-callback
and the logic here that pushes the response before the XHR object: https://github.com/KartikTalwar/gmail.js/blob/c18173c0305736ce57f016977097aa45fecd34b4/src/gmail.js#L2314


josteink:
Checked and looks good to me. Thanks for helping out!

josteink:
Published on npmjs as v1.0.20!

PR #683, opened 2022-04-27
Mark identifier on email_data.new.get as optional

cancan101:
Makes the ts signature consistent with both the docs an the code itself: https://github.com/KartikTalwar/gmail.js/blob/c18173c0305736ce57f016977097aa45fecd34b4/src/gmail.js#L4405-L4410


josteink (approved):
LGTM!

Issue #686, opened 2022-05-16
API to Add Items to "More Send Options" Menu


cancan101:
Add an API to add options to the "More Send Options" Menu which is exposed by clicking the arrow on the right side of the send button: 
<img width="191" alt="image" src="https://user-images.githubusercontent.com/51059/168646802-7dc434ec-0b0e-4f51-bab7-ca0295979fbf.png">

josteink:
If doable, it does sound useful. PRs welcome!

cancan101:
this seems to work:
```ts
(gmail as any).tools.add_more_send_option = function (
  composeWindow: GmailDomCompose,
  content_html: string,
  onClickFunction: Function,
  styleClass?: string
): JQuery {
  var div = $(document.createElement('div'));
  div.attr('class', 'J-N yr');
  div.attr('style', 'user-select: none;');
  div.attr('role', 'menuitem');

  var button = $(document.createElement('div'));
  var buttonClasses = 'J-N-Jz ';
  if (styleClass !== undefined) {
    buttonClasses += styleClass;
  }
  button.attr('class', buttonClasses);
  button.attr('style', 'user-select: none;');
  button.html(content_html);
  button.click(onClickFunction as any);

  div.append(button);

  var scheduledSend = composeWindow.find('.J-N.yr').last();
  div.insertAfter(scheduledSend);
  console.log(composeWindow, scheduledSend);

  return button;
};
```

though I think there might be a better API passing in the imgSrc and the text since right now the HTML passed in looks like:
```html
<img class="v5 J-N-JX" role="menuitem" src="images/cleardot.gif" alt="" style="user-select: none;">Text
```

cancan101:
and here is the draft PR: https://github.com/KartikTalwar/gmail.js/pull/687

PR #687, opened 2022-05-17
Add add_more_send_option

cancan101:
Closes: https://github.com/KartikTalwar/gmail.js/issues/686

### DIFF
--- a/src/gmail.d.ts
+++ b/src/gmail.d.ts
@@ -711,9 +711,10 @@ interface GmailTools {
     add_compose_button(composeWindow: GmailDomCompose, content_html: string, onClickFunction: Function, styleClass?: string): JQuery;
     add_more_send_option(
       composeWindow: GmailDomCompose,
-      content_html: string,
+      buttonText: string,
       onClickFunction: (event: JQuery.Event) => void,
-      styleClass?: string
+      styleClass?: string | undefined,
+      imgSrc?: string
     ): JQuery;
     /**
        adds a button to an email attachment.
--- a/src/gmail.d.ts
+++ b/src/gmail.d.ts
@@ -712,13 +712,7 @@ interface GmailTools {
     add_more_send_option(
       composeWindow: GmailDomCompose,
       content_html: string,
-      onClickFunction: JQuery.TypeEventHandler<
-        HTMLDivElement,
-        null,
-        HTMLDivElement,
-        HTMLDivElement,
-        'click'
-      >,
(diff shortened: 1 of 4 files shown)


src/gmail.d.ts
```diff
@@ -708,7 +708,19 @@ interface GmailTools {
...
+    add_more_send_option(
+      composeWindow: GmailDomCompose,
+      content_html: string,
+      onClickFunction: JQuery.TypeEventHandler<
+        HTMLDivElement,
+        null,
+        HTMLDivElement,
+        HTMLDivElement,
+        'click'
+      >,
```
josteink, line 721: I think this type is more complicated than it needs to be.

What we pass in as an on-click function should just be an anonymous `() => void` or `Function`.

cancan101: Yes, I can simplify though the type is actually `(event: JQuery.Event) => void`

josteink: I'd prefer it to be either be unwrapped to the DOM-native, or just a simple `() => void`, just like the rest of the `onClickHandlers` we have all around :)

cancan101: Unwrapping would be a semantic change rather than just a tweaking of the declared types. i.e. for `add_compose_button`, the `onClickFunction` is passed directly to jquery's `click`: https://github.com/KartikTalwar/gmail.js/blob/b7e5b814e48de270fd33a04f85b4682bb7676892/src/gmail.js#L3590

unwrapping to DOM-native would be a breaking change to the code it self. I think for consistency with other methods, the `onClickFunction` here should match. Further the return type for all of the `add_XXX_button` is `JQuery` rather than the `HTMLElement`. It seems reasonable that we are taking in and return JQuery wrapper rather than native DOM. 

Similarly, for other methods, `onClickFunction` is declared as `Function`:
https://github.com/KartikTalwar/gmail.js/blob/b7e5b814e48de270fd33a04f85b4682bb7676892/src/gmail.d.ts#L709-L711
`(event: JQuery.Event) => void` is just a more precise type than currently specified.

cancan101: Ok, I can change it to `Function`. That being said, I think if I were to make the function `(event: JQuery.Event) => void`, users can still pass in a `() => void` i.e. that would not be a breaking change as the new type is more permissive.

josteink:
As a final request, could you please add this new function to the README, so that it’s documented? 🙂

josteink:
Great! Looks good to me. Thanks for hanging in there. 

Consider it merged!

PR #689, opened 2022-05-22
lookup is optional for compose.dom()

lirikooda:
https://github.com/KartikTalwar/gmail.js/blob/b7e5b814e48de270fd33a04f85b4682bb7676892/src/gmail.js#L4075-L4077


josteink:
You’re absolutely right. Good catch!

PR #690, opened 2022-05-22
Improved compose.recipients() typing

lirikooda:
`recipients` is expecting a `type` or `flat`, but not both. Also return type is different according to args, and not some internal logic. Therefore, its better to overload.
Note this is a breaking change: someone might have used the method with both `type` and `flat` (which would result in the latter being ignored), but I think since breaking is only for typing, its not a blocker.


josteink:
You’re absolutely right. Seconding this change! 🙂

Issue #148, opened 2015-03-02
Set to: cc: bcc: on api.dom.compose object


lievenjanssen:
In the code there is a comment on the to: cc: and bcc: fields (TODO: ability to set)

Are there any plans to implement the to: cc: bcc: in the near future?

brentkelly:
@lievenjanssen - I had a brief look at this but didn't need it for my app so haven't implemented it yet.

It is more complex than the subject etc because it creates a bunch of DOM objects for each to/cc/bcc address to allow you to delete it with a click, drag it around etc. IIRC after a bunch of experimentation I found a hidden element that stored all to addresses, and that if you updated that & then clicked off it, gmail would add in all the DOM elements. So if that is correct then the best approach would be to update that field & then simulate a click. I had a bit of trouble simulating that click with jQuery but it looks like others have had more luck in other threads - can't remember the threads off the top of my head but I think they just used standard js to execute a mousedown followed by a mouseup?

Hope that helps.

brentkelly:
@lievenjanssen - OK the mouse event thread I was referring to is #117 which seems to work fine for triggering the search.

I tried this approach here but it doesn't seem to work for me. I don't have much experience at simulating click events some perhaps someone knows more.

```javascript

// set email textarea field
$('textarea[name=to]').val('my@email.com');

// try to simulate a click on the subject input - doesn't work
var mouseUp = document.createEvent('MouseEvents');
mouseUp.initEvent( 'click', true, false );
$('input[name=subjectbox]')[0].dispatchEvent(mouseUp);

// try give focus to the subject input - doesn't work
$('input[name=subjectbox]').focus();
```

Aside from that approach, I'm sure really how you could implement this functionality without re-implementing the Gmail interface by recreate the entire DOM tree - including click & drag events - that Gmail creates when you add a to address. Unless Gmail is nice enough to apply those events automatically to all elements with a specific class name. I haven't tested that but I'd be very surprised, and even if it did, that would be some seriously brittle code :).

glebovilya:
It seems like there have to be several places to add needed data - not only input[name="to"]. Probably solution is to add:

1) <div class="vR"><span class="vN vP a3q" email="email@email.com"><div class="vT">email@email.com</div><div class="vM"></div></span><input name="to" type="hidden" value="email@email.com"></div>

have to be added as previous sibling of textarea

2)<div id=":xx" class="oL aDm az9"><span email="email@email.com" class="aDr">email@email.com</span></div>

[...]

PR #691, opened 2022-05-22
Add/remove recipient from compose

lirikooda:
See also #148
I'd love to add unit tests, but would need some pointers for that please.

### DIFF
--- a/src/gmail.d.ts
+++ b/src/gmail.d.ts
@@ -54,11 +54,9 @@ declare type GmailPageType =
    */
 declare type GmailEmailAddress = string[];
 
-declare type GmailDomComposeRecipients = {
-   to: string[];
-   cc: string[];
-   bcc: string[];
-}
+declare type GmailDomComposeRecipientField = 'to' | 'cc' | 'bcc'
+
+declare type GmailDomComposeRecipients = Record<GmailDomComposeRecipientField, string[]>
 
 declare type GmailAttachmentDetails = {
     attachment_id: string,
@@ -80,11 +78,8 @@ declare type GmailEmailDetails = {
     attachments_details: GmailAttachmentDetails[],
     subject: string,
     content_html: string,
-    content_plain: string,
-    to: string[],
-    cc: string[],
-    bcc: string[]
-};
+    content_plain: string
+} & GmailDomComposeRecipients;
 
 declare type GmailEmailData = {
     thread_id: string,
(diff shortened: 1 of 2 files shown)


josteink:
Hey there. 

While the changes here don’t look *wrong* per se, and the TypeScript definitions are definitely improved, it seems like the new functions added are kind of overlapping with what the `gmail.dom.email.{to,cc,bcc}` functions are supposed to provide. 

Are there any reason those functions don’t work for you as is?

lirikooda:
I agree that it'd be better to have the add/remove functionality work with the current getter api.
However, it simply doesn't work as is. See the relevant issue: https://github.com/KartikTalwar/gmail.js/issues/148
I could reopen it and expand there.
In any case, the returned value from `{to, cc, bcc}` is simply not enough to make the functionality work. Especially "remove" which is unrelated to the textarea at all.

josteink:
Right. But if they don’t work as is, wouldn’t it make more sense to fix them in place, rather than adding new implementations in new functions?

I agree we lack a good/consistent remove-api though. 

On a side-note (it’s early Sunday here) why are modifying api.dom.email instances anyway? They’ve long been sent and any change we do is purely in the UI of the user, right? 🙂

lirikooda:
I think fixing in place is kinda problematic. At the moment, these methods accept a single string, you'd expect a `string[]` if you want to set entire recipient field. And I wouldn't want to break functionality.
Adding ui changes is very useful, having it as part of the library is an asset IMO.
In any case, I'm more than happy to align with the way we'd want the apis to be defined, let's just keep in mind all the use cases.

josteink:
Sorry about the slow response here. Work has taken priority lately :)

So to recap:

* The PR title says add/remove from compose
* The code says add/remove from email
* This has caused a little confusion for this poor maintainer as he is struggling with context-switches :smile: 
* We already have some methods, at least for adding addresses to emails.
* But you don't want to risk breaking those by adding support for your use-case.

That's about correct?

I appreciate how you want to avoid breaking existing code, but I'd also don't want to make the API difficult to navigate by having several similar, but not quite, methods for doing kinda the same stuff, but in different ways. Not to mention the code-duplication that might lead to. So yeah.

In general, I there's no problem making a method kinda polymorphic. In fact we do that all over in `gmailjs` already :grin: 

For instance `gmail.dom.email.to()` accepts both arrays and non-array inputs. There's no `gmail.dom.email.cc()` or `gmail.dom.email.bcc()`, but I think adding them would make the API more at parity with the `gmail.dom.compose`-class, so that would be a good thing.

So if that's ok, that would leave the only thing remaining to be removing recipients.

That can be solved through a seperate method, or an optional bool like

````javascript
    const email = gmail.dom.email();
    email.to({name: "John Doe", email: "john.doe@gmail.com"});
    const remove = true;
    email.to({name: "John Doe", email: "john.doe@gmail.com"}, remove);
````

How about that?

lirikooda:
Sounds good. Sorry I wasn't too familiar with the difference between compose/email contexts.
I'll work on adjusting the api and will reopen the PR then.

josteink:
Sounds good. Let me know if you want or need any more feedback 🙂

PR #692, opened 2022-05-22
Added explicit compose type function

lirikooda:
callback returning compose type is now redundant, but keeping it for compatibility.


josteink:
Change looks good, but it’s a new public function. 

Please add that function to the documentation in README as well 🙂

lirikooda:
> Change looks good, but it’s a new public function.
> 
> Please add that function to the documentation in README as well 🙂

Thanks, added!

josteink:
Great. Thanks for the contributions! Merged!

PR #695, opened 2022-05-31
Fix insertion_observer accessing state two different ways

cancan101:
`dom_observer_map` is already passed in and accessed within the function


josteink (approved):

josteink:
Change looks correct and I'll be glad to merge it.

But is there any reason you're implementing this change apart from simply being .. uhm simpler? Does it fix anything? :smile:

PR #698, opened 2022-05-31
Fix typing for the register args

cancan101:


josteink:
On the surface of things, this change looks good. I must admit though that this is a part of the code-base where I'm not *that* well versed, and there might be nuances in how this function is used which I'm not aware of.

But since we're only updating the *typing*, and from what I can tell, the typing-change does not restrict the way the function can be used beyond what's clearly correct...

Let's get this merged :smile:

josteink (approved):

PR #700, opened 2022-05-31
Create generic concept for GmailBindAction to allow extension

cancan101:


src/gmail.d.ts
```diff
@@ -863,7 +863,7 @@ interface GmailObserve {
       trigger the related event and fire off any relevant bound callbacks
```
cancan101, line 863: Register could be changed to:
```ts
register(action: T, args: string | StringDict): void;
````
but that would be a breaking change

**Comment by josteink** [comment]
The current solution looks good to me. To be honest, I'd probably only apply a `string | GmailBindAction` and call it day, but this is even better.

As for the code, couldn't/shouldn't the generic parameter rather be declared on the method instead of the class, since it only applies to that method?

Otherwise PR looks good to me.

**Comment by josteink** [comment]
> Register could be changed to:
>
>     register(action: T, args: string | StringDict): void;
>
> but that would be a breaking change

I don't think that would be correct, would it?

`StringDict` would only accept (non-key) values of type `string`, but `handler` needs to be a function, right?

**Comment by cancan101** [comment]
My point there as on the change of the type of `action`, i.e:
```ts
register(action: string, args: string | StringDict): void;
```
to
```ts
register(action: T, args: string | StringDict): void;
```

the type of `args` was recently changed in another PR of mine

cancan101:
At present this code does not work:
```ts
const GmailFactory = require('gmail-js');
const gmail = new GmailFactory.Gmail(jQuery) as Gmail;
gmail.observe.register('popout_thread', popoutThreadConfig);

// this is an issue as on_dom expects a GmailBindAction
gmail.observe.on_dom(
  'popout_thread',
  ...
)
```

if you just change `on_dom` to take a string then you weaken the type safety that exists in cases w/o a custom dom event.

josteink:
If the TS compiler can infer T for the main Gmails-class, that would be a nice feature to working. 

If not, it’s definitely reduces the ergonomics for pretty much all other use-cases, and I’m against that.

cancan101:
well the idea is to provide a default for T so in 99% of use cases the users can use as is. Only in the case of wanting to register custom dom events does the user need to specify the typing.

josteink:
I’d be OK with that.

josteink:
Just checking in. Did you get anywhere with this one? :)

cancan101:
I think good for review. Only question is whether it makes sense to tighten the typing on register per my comment: https://github.com/KartikTalwar/gmail.js/pull/700/files#r886143560

src/gmail.d.ts
cancan101: Updated slightly given recent changes:
```ts
register(action: T, args: string | DomObserverConfig): void;
```
which would limit the values of action to those that are pre-declared.

josteink:
Back from vacation now. Will try to take a look at this when I have time.

src/gmail.d.ts
```diff
@@ -870,7 +870,7 @@ interface GmailObserve {
...
       This function should return true if a dom observer is found for the specified action
      */
-    on_dom(action: GmailBindAction, callback: Function): void;
+    on_dom(action: GmailBindAction | T, callback: Function): boolean;
```
josteink, line 873: Coming back to this PR and looking into it... Why do we need to change the return-type?

Does anyone use it for anything? :smile:

josteink:
Apart from some minor nitpicks about return-types, this looks OK to me.

 I won't need it, but if it doesn't otherwise interfere with "normal" GMailJS usage, I don't see the harm either.

src/gmail.d.ts
cancan101: I just updated the return type declaration here to be consistent with the docs on the function which says it returns a boolean.

josteink: Fair enough. Let’s make it so!

josteink:
Ready when you are 🙂

josteink:
Any news on this one? It seems like it kinda stopped :smile:

cancan101:
This should be ready for review. This PR should have no breaking changes. There is an option to tighten the types on `register` but that would be a breaking change (requiring users to opt in to this new typing feature)

josteink (approved):
LGTM!

PR #701, opened 2022-06-07
Use class to control the image for more send option

cancan101:
This is consistent with existing item(s) in the menu

Fixes some issues with https://github.com/KartikTalwar/gmail.js/pull/687


josteink:
It may or may not be typical JS conventions, but in GmailJS we try to keep and enforce double-quotes, not single-quotes.

Could you update the method to use this instead of the single-quotes you are using now? :)

josteink (approved):
LGTM

PR #703, opened 2022-06-21
Fixed subject being returned instead of body when Mixmax is installed

DrParanoia:
The Mixmax extension for Gmail is changing the `subject` field from `input` to `contenteditable`, causing the `body` selector to fetch the wrong thing.

Added an extra check, to make sure we are not fetching the `subject`.


josteink:
Looks good to me. Consider it merged 🙂

josteink:
This is now published on npmjs as `v1.0.22`. Thanks for contributing!

Issue #707, opened 2022-07-15
get.loggedin_accounts() is gone without replacement


onestep:
Latest release of gmail.js has removed functions, which could have been used in some projects:
* get.loggedin_accounts()
* get.delegated_to_email()
* helper.get.is_delegated_inbox()

They can be supported in new Gmail as well.

josteink:
Should be fixed in next, upcoming release thanks to you :smile:

PR #708, opened 2022-07-15
fix: update code in get.loggedin_accounts() and helper.get.is_delegat…

onestep:
…ed_inbox() to support new Gmail (#707)

### DIFF
--- a/src/gmail.d.ts
+++ b/src/gmail.d.ts
@@ -106,6 +106,12 @@ declare type GmailLastActive = {
     time_relative: string
 };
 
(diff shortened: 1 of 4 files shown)


onestep:
Hello, thanks for the comments, I've added additional commit to address them. 🙂 I just wanted to keep existing style and minimize diff as much as possible.

josteink:
Tested and works fine. Consider it merged!

josteink:
v1.0.24 out on NPMJs now, with these fixes included :)

PR #711, opened 2022-07-20
fix: corrected outdated type definitions (fix #710)

### LINKED ISSUE #710: Inconsistencies in type definitions
Some functions are incorrectly typed, i.e.:
* `GmailDomCompose` `to`, `cc,` `bcc` functions accept value, which would be set to the corresponding DOM element;
* `GmailDomEmail` `dom` function could be called with no lookup;
* `check,is_peoplekit_compose` `composeElement` parameter could be of type `JQuery` as well as `HTMLElement`.

Type definitions should be updated to allow mentioned cases.

onestep:


josteink (approved):

josteink:
On vacation without laptop, so I can merge based on it looking good. I can’t however make a new release just yet 😉

onestep:
Thank you! I'd better wait for #708 to hit the new release as well. 🙂

josteink:
v1.0.24 out on NPMJs now, with these fixes included :)

Issue #713, opened 2022-08-03
compose.recipients() does not work when PeopleKit suggestions are displayed


onestep:
**Steps to reproduce:**
1. open Gmail for PeopleKit-enabled account
2. open new compose window, add some recipients you used in previous emails and wait for suggestion (unfortunately I don't know the exact way to force suggestions - most likely if you frequently send emails to the same three contacts, when you add two in the new compose, Gmail will suggest third)
![image](https://user-images.githubusercontent.com/1419318/182607281-dca2af9d-b702-491d-8f82-30dae94b38db.png)
3. try to read `gmail.dom.composes()[0].recipients()`

**Expected result:**
* object with recipients info is returned

**Actual result:**
![image](https://user-images.githubusercontent.com/1419318/182607781-7eca084f-4e66-4d5b-ba8c-34e86953509c.png)

josteink:
Closed via https://github.com/KartikTalwar/gmail.js/pull/714.

PR #714, opened 2022-08-03
fix: add [name] selector for all PeopleKit recipient containers (#713)

onestep:
- avoid using closest() if type if specified in options

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -3955,14 +3955,14 @@ var Gmail = function(localJQuery) {
            options.flat  boolean if true will just return an array of all recipients instead of splitting out into to, cc, and bcc
         */
         recipients: function(options) {
-            if( typeof options !== "object" ) options = {};
+            if (typeof options !== "object") options = {};
             const peopleKit = api.check.is_peoplekit_compose(this.$el);
 
             const type_selector = options.type ? "[name=" + options.type + "]" : "";
 
             const found = peopleKit ?
-                this.$el.find("tr.bzf " + type_selector + " div[data-hovercard-id]").map((_, el) => ({
-                    type: el.closest("div[name]").getAttribute("name"),
+                this.$el.find("tr.bzf " + (type_selector || "div[name]") + " div[data-hovercard-id]").map((_, el) => ({
+                    type: options.type || el.closest("div[name]").getAttribute("name"),
                     email: el.getAttribute("data-hovercard-id")
                 })) :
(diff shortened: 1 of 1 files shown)


src/gmail.js
```diff
@@ -3955,14 +3955,14 @@ var Gmail = function(localJQuery) {
...
-            if( typeof options !== "object" ) options = {};
+            if (typeof options !== "object") options = {};
             const peopleKit = api.check.is_peoplekit_compose(this.$el);
 
             const type_selector = options.type ? "[name=" + options.type + "]" : "";
 
             const found = peopleKit ?
-                this.$el.find("tr.bzf " + type_selector + " div[data-hovercard-id]").map((_, el) => ({
-                    type: el.closest("div[name]").getAttribute("name"),
+                this.$el.find("tr.bzf " + (type_selector || "div[name]") + " div[data-hovercard-id]").map((_, el) => ({
```
josteink, line 3964: Wouldn't it be better to make `"div[name]"` the new default for `type_selector`  (line 3961) when no type is provided? Because that's effectively what this change does, right?

It would make this code a little bit easier/cleaner to read.

src/gmail.js
```diff
@@ -3955,14 +3955,14 @@ var Gmail = function(localJQuery) {
...
+                    type: options.type || el.closest("div[name]").getAttribute("name"),
```
josteink, line 3965: Could another option here be to reverse this statement, using a null coalescing operator?

`````js
type: el.closest("div[name]")?.getAttribute("name") || options.type,
`````

Or even more: Is this change really needed at all? with the previous fix? :)

src/gmail.js
onestep: I would agree if there would be no `is_peoplekit_compose` check. In other case, there is `<input name="to">`, and this selector is appended as a part of input selector (not as a parent).

src/gmail.js
onestep: This change allows us to avoid traversing the DOM, therefore providing result faster. If we use specific `"div[name=to]"` parent selector, there is no need to look up for the parent, and then taking its `name` attribute. 🙂

josteink: Fair enough!

src/gmail.js
josteink: Right. I see. Let’s keep this as is too then.

josteink:
Looks good to me. Consider it merged.

josteink (approved):

PR #723, opened 2022-09-13
fix: detect and parse send message payload, fixes #722

### LINKED ISSUE #722: send_message event stopped working
It is still something to be confirmed, but I can see that both `send_message` and `send_scheduled_message` does not work anymore.

These hooks are now never triggered:
```
this.gmail.observe.before("send_message", (url: string, body: string, data: Mail) => {});
this.gmail.observe.before("send_scheduled_message", (url: string, _body: string, data: Mail) => {});
```

### Browser differences

- Works fine in Firefox
- Does not work in Chrome (with the same Google account)

I can see there is different parameters passed in URL - Chrome v105

image

Firefox v104
image

I am trying to investigate if it is payload or something...

huksley:
I have code for both versions now (old style and new), but it is difficult for me to validate old APIs working because my Firefox also receives new version now.

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -672,12 +672,8 @@ var Gmail = function(localJQuery) {
     };
 
 
-    api.tools.error = function(str) {
-        if (console) {
-            console.error(str);
-        } else {
-            throw(str);
-        }
+    api.tools.error = function(str, ...args) {
+        console.error(str, ...args);
     };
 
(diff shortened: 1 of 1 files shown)


josteink:
This is all terrible and we both know it :laughing: 

That said, if this fixes some real issues, I'm all for getting them merged.

PR #726, opened 2022-09-22
Fix subject parsing in parse_sent_message_payload_new

### LINKED ISSUE #725: the value of subject in the data passed into processSend is not correct
I had to use:
```ts
const bodyData = JSON.parse(body);
const emailSubject = bodyData[1]?.[0]?.[0]?.[1]?.[1]?.[13]?.[0]?.[7];
```

It looks like the current path being returned is: `[1][0][0][1][1][2][0][4][0][8]`
which is from:
https://github.com/huksley/gmail-js/blob/79a18883343a55002f9010bca4dbca02d3349297/src/gmail.js#L1790

It looks like the subject line is contained in the parent object to what is passed to `parse_sent_message_payload_new`.

cancan101:
Closes #725

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -1787,7 +1787,7 @@ var Gmail = function(localJQuery) {
 
             const sent_email_id = sent_email[0];
 
-            const sent_email_subject = sent_email[8];
+            const sent_email_subject = sent_email[7];
             const sent_email_timestamp = sent_email[6];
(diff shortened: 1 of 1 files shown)


josteink:
Thanks! LGTM. Merged!

I’ll try to get this released as a new version on npmjs later on.

josteink:
Now pushed to npmjs as `v1.1.4`!

Issue #729, opened 2022-10-12
Attachments from fd_embedded_json are not correctly parsed after Gmail update


onestep:
After the Gmail internal API update I've noticed that attachments are not correctly parsed due to the exception:
![image](https://user-images.githubusercontent.com/1419318/195415352-d85bb2a1-6949-4955-8acf-5082c00fabe0.png)

`parse_fd_embedded_json_attachments()` should be updated to use new property indexes.

MadcowD:
I'm also having this error.

josteink:
A new version with a fix for this issue has now been released as version `1.1.5`.

PR #730, opened 2022-10-12
fix: update property indexes in parse_fd_embedded_json_attachments() …

onestep:
…(#729)


onestep:
Sorry for direct editing of test input, but it did not contain attachments in any thread, so I've dragged in a piece of raw data from one of my test accounts.

josteink:
Looks good to me.

Add the fix to the changelog and bump the version too? :smile:

Issue #731, opened 2022-10-17
gmail.observe.on('view_thread') not working


ZayaSerg:

josteink:
More info would be appreciated 🙂

cancan101:
As of somewhat recently, a subscription to the `view_thread` event no longer seems to fire even though a thread is viewed. I will have to investigate further.

cancan101:
I changed the `sub_selector` to: `'div.PeIF1d'` and it works

josteink:
PRs welcome.

josteink:
Agreed. 

If this is going to be a recurring problem, adding general support for multiple sub-selectors is probably not a bad and nothing I fundamentally have an issue with. 

Try it out and see if you can make it work?

MadcowD:
This looks good!

MadcowD:
@josteink is there anyway you could merge this? I'd love to update our app.

josteink:
I was too quick in the review/testing of previous PR.

Upon proper testing I can't make this work at all. Not on new PR-code nor on old master code.

I guess that means there's more work to be done here, but I can't fix that now. I don't have time, at least until sometime next week.

cancan101:
@josteink I am a bit confused here. Are you saying this change  _broke_ your experience?

josteink:
No I'm saying it didn't work without this fix, but it doesn't work with it either. 

view_email event works though (no change).

josteink:
The fix in its current state is released in new version `1.1.5`. No guarantees it will work, but at least it should make it easier to test.

cancan101:
Okay, I just did a hard refresh of my gmail and it seems like the class name used by gmail has changed once again. So agreed, this fix does not work.

josteink:
I guess that means we might have to use a higher level selector (.bu?) and try to filter it more intelligently in code further down (look for data-thread-id properties)?

Basically something which is generic enough to survive such changes, but not wide enough to cause performance issues.

cancan101:
Ok, I think a better sub_selector is `iY` which given its length and proximity to `if` is more likely to be stable (?).

josteink:
PRs welcome 🙂

cancan101:
https://github.com/KartikTalwar/gmail.js/pull/735

josteink:
Merged, available in 1.1.6.

PR #732, opened 2022-10-18
Fix sub_selector used for view_thread

cancan101:
Closes https://github.com/KartikTalwar/gmail.js/issues/731

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -280,7 +280,9 @@ var Gmail = function(localJQuery) {
     };
 
     api.check.is_thread = function() {
+        // this should match the sub_selector (nH -> if)
         var check_1 = $(".nH .if").children(":eq(1)").children().children(":eq(1)").children();
+        // Bu ->
         var check_2 = api.get.email_ids();
 
         return check_1.length > 1 || check_2.length > 1;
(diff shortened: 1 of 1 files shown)


src/gmail.js
```diff
@@ -280,7 +280,9 @@ var Gmail = function(localJQuery) {
     };
 
     api.check.is_thread = function() {
+        // this should match the sub_selector (nH -> if)
         var check_1 = $(".nH .if").children(":eq(1)").children().children(":eq(1)").children();
+        // Bu ->
```
josteink, line 285: We probably don't want these comments here to be part of this PR.

josteink (changes requested):

src/gmail.js
cancan101: These comments were intentional as they relate back to how the selector/sub selector values are used in multiple places in the code. I had to spend some time understanding the coupling.

cancan101: Yea happy to make this more clear. Basically the `Bu` was one of the selectors and `nH` is the other. There are two "code" paths to flag this as a thread based on the selector.

cancan101: I can add something along the lines of this comment ^ if you think that will make it clear

josteink: This looks good now.

josteink:
Code change looks OK to me.

As for testing, I can't reproduce the original issue, but I can confirm that this code still seems to work for me none the less, so it shouldn't have caused any regressions at least.

Consider it merged!

josteink (approved):

Issue #766, opened 2023-10-10
Synchronous 'DOMNodeInserted' DOM Mutation Event is deprecated


cancan101:
Seeing this warning from Chrome:
```
Listener added for a synchronous 'DOMNodeInserted' DOM Mutation Event. This event type is deprecated (https://w3c.github.io/uievents/#legacy-event-types) and work is underway to remove it from this browser. Usage of this event listener will cause performance issues today, and represents a risk of future incompatibility. Consider using MutationObserver instead.
```
which I think is coming from this line: https://github.com/KartikTalwar/gmail.js/blob/080b39455acba8620efbb50a7caf984af8fbdc86/src/gmail.js#L2678

there was some discussion about the implementation:
- https://github.com/KartikTalwar/gmail.js/issues/560
- https://github.com/KartikTalwar/gmail.js/issues/636#issuecomment-777347684

but it may be the case that this needs to be changed sooner than later.

josteink:
Thats probably very likely yes. Thanks for the heads up.

I'm going to assume we can fix this somehow by making a  non-synchronous handler instead (from the browsers POV). For my extension that's probably fine. I'm going to assume most people using our `view_email` events (and friends) don't have a hard-dependency on it being synchronous either.

Does anyone have time to look into what the alternate implementation looks like?

Kravimir:
The alternative is to use a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver).

cancan101:
See https://developer.chrome.com/blog/mutation-events-deprecation/#use-mutationobserver-instead for migration suggestion.

I would think these might be able to be merged unless there is a reason one uses `body` and one uses `document`: https://github.com/KartikTalwar/gmail.js/blob/080b39455acba8620efbb50a7caf984af8fbdc86/src/gmail.js#L2678C33-L2702

Issue #765, opened 2023-09-18
gmail.observe.on('view_email' is not fired


ionh:
have this code from the boilerplate but looks like now `view_email` is not fired anymore? Other events like `view_thread` or `compose` are working

```
gmail.observe.on("load", () => {
    const userEmail = gmail.get.user_email();
    console.log("Hello, " + userEmail + ". This is your extension talking!");

    gmail.observe.on("view_email", (domEmail) => {
        console.log("Looking at email:", domEmail);
        const emailData = gmail.new.get.email_data(domEmail);
        console.log("Email data:", emailData);
    });

    gmail.observe.on("compose", (compose) => {
        console.log("New compose window is opened!", compose);
    });
});
```

Any ideas?

josteink:
Hey there and thanks for the bug-report.

I've tested and I can reproduce. I've given this a little run in the debugger and it seems to be timing-related.

That is, our selectors are correct, but we are trying to look for a HTML element which has not yet been inserted.

There may be different (and better) ways to fix this, but I managed to get the event triggering by making this small change:

````diff
                 // class depends if is_preview_pane - Bu for preview pane, nH for standard view,
                 // FIXME: the empty class ("") is for emails opened after thread is rendered (causes a storm of updates)
                 class: ["Bu", "nH", ""],
-                sub_selector: "div.adn",
                 handler: function(match, callback) {
-                    match = new api.dom.email(match);
-                    callback(match);
+                    setTimeout(() => {
+                        match = match.find("div.adn.ads");
+                        if (match.length) {
+                            match = new api.dom.email(match);
+                            callback(match);
+                        }
+                    }, 0);
                 }
             },
````

This probably should be looked into a bit deeper, and a new version will need to be released.

Do you have the ability to help out testing the above patch on your end?

**Comment by bchenhs** [comment]
We ran into this earlier and found a workaround. There are two parts to the fix:
1. Insert the setTimeout to yield and let Gmail view to render. We placed it [here](https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js#L2675) where `insertion_observer` is invoked.
2. Update the [insertion_observer implementation](https://github.com/KartikTalwar/gmail.js/blob/master/src/gmail.js#L2781) to handle each message view elements:

```
if (element.length) {
  let idx;
  for (idx = 0; idx < element.length; idx++) {
    var handler = config.handler ? config.handler : function(match, callback) { callback(match); };
[...]

PR #767, opened 2023-10-10
use MutationObserver to track dom node additions.

cancan101:
Closes https://github.com/KartikTalwar/gmail.js/issues/766

CC @Kravimir @josteink

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -2672,13 +2672,6 @@ var Gmail = function(localJQuery) {
                 api.tracker.observing_dom = true;
                 //api.tracker.dom_watchdog = {}; // store passed observer callbacks for different DOM events
 
-                // this listener will check every element inserted into the DOM
-                // for specified classes (as defined in api.tracker.dom_observers above) which indicate
-                // related actions which need triggering
-                $(window.document).on("DOMNodeInserted", function(e) {
-                    api.tools.insertion_observer(e.target, api.tracker.dom_observers, api.tracker.dom_observer_map);
-                });
-
                 // recipient_change also needs to listen to removals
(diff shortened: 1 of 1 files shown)


josteink:
Hey @cancan101 and thanks for the super-quick pick!

The change looks simple enough. I've also tested with my own extension, everything works, and from what I can tell there are no adverse effects.

I'd be happy to merge this change. How about about you @Kravimir?

josteink:
No feedback from @Kravimir in roughly a week... So I say this is good to merge.

Kravimir:
Not necessarily. In my case it often means I haven't remembered to come back to look at it.

However, in this case, the change is good, although I do note that the observed node is changed from "document" to "document.body".

josteink (approved):

josteink:
> Not necessarily. In my case it often means I haven't remembered to come back to look at it.

I know such things happen. It happens to me too!

But that often mean an issue isn't *that* critical to the person involved. In those cases, if others feel a need to move on, it should be OK to do so, especially if notifying about that happening 😄 

> However, in this case, the change is good, although I do note that the observed node is changed from "document" to "document.body".

Nice observation. Looking into docs here, I don't think this is going to be an issue for our use.

We're never going to need to inspect for changes in the `<HEAD>` element are we?

cancan101:
so, can this PR be merged?

josteink:
Yeah. Sure thing. Sorry about the delay!

cancan101:
Can we can a release cut with this PR? I think `gmail.observe.on('view_thread')...` may be broken without it. See: https://github.com/KartikTalwar/gmail.js/issues/765

PR #771, opened 2023-11-22
New versions of gmail don't have that table anymore

MadcowD:
Feel free to discuss


josteink:

I've tested the existing version on my free grandfathered "legacy" GSuite account, and the existing code seems to work fine there, at least for me.

Maybe the best approach would be to combine the two checks into one, just to be sure it keeps on working equally well for everyone? Could you update your PR to do that instead?

If you do that, I'll merge and push a new version to npmjs.

josteink:
Ive retested, and now this code is the only one which works on plain gmail.com accounts and my grandfathered Google apps account. Consider this good enough :)

PR #776, opened 2024-02-08
Fix the navigation_count function and selector

stevenirby:
This fixes two of the methods `api.helper.get.navigation_count` and `api.get.storage_info`. It returns a fair amount of data back:

```
{storage_info: 
{used: 14.32, total: 15, percent: 95},
unread_draft_emails: 152
unread_emails: {inbox: 882, drafts: 152, spam: 0, forum: 0, update: 0, …}
unread_forum_emails: 0
unread_inbox_emails: 882
unread_promotion_emails: 7
unread_social_emails: 0
unread_spam_emails: 0
unread_update_emails: 0}
```

I'm trying to target the aria labels and accessibility tags, as I think it's a better move to try to parse the page like https://testing-library.com/ library does testing. Looking at the page like a real user or a screen reader would, rather than targeting class names or such, which is a cat-and-mouse game.

I'm also doing console.warns when things aren't found for easier debugging. Throwing an error or console.error seemed a bit too much.

I'm totally open to any advice or feedback!

### DIFF
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -636,17 +636,17 @@ var Gmail = function(localJQuery) {
 
     api.helper.get.navigation_count = function(i18nName) {
         const title = api.tools.i18n(i18nName);
-        const dom = $("div[role=navigation]").find("[title*='" + title + "']");
+        const dom = $('[aria-label*="' + title + '"]');
 
-        if (dom || dom.length > 0) {
-            // this check should implicitly always be true, but better safe than sorry?
-            if(dom[0].title.indexOf(title) !== -1) {
-                const value = parseInt(dom[0].attributes['aria-label'].value.replace(/[^0-9]/g, ""));
-                if (!isNaN(value)) {
-                    return value;
-                }
+        if (dom && dom.length > 0) {
+            const label = dom[0].attributes['aria-label'];
+            const value = parseInt(label.value.replace(/[^0-9]/g, ""));
+            if (!isNaN(value)) {
+                return value;
             }
         }
+
+        console.warn("Gmail.js: Unable to find the count for " + i18nName);
 
         return 0;
     };
--- a/src/gmail.js
+++ b/src/gmail.js
@@ -118,7 +118,42 @@ var Gmail = function(localJQuery) {
 
 
     api.get.user_email = function() {
-        return api.tracker.globals[10];
+        const emailRegex = /\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b/;
+
+        function extractEmail(textContent) {
+            const matches = textContent.match(emailRegex);
+            return matches && matches.length > 0 ? matches[0] : undefined;
+        }
+
+        try {
+            // First try to get the email from an element with an aria-label
+            const label = document.querySelector('[aria-label^="Google Account"]');
+            if (label) {
+                const email = extractEmail(label.textContent);
+                if (email) {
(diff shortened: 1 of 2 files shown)


josteink:
Ive given the PR a test-run on a few different accounts:

* Regular gmail.com account
* OG, grandfathered "Google Apps" account
* Paid, company GSuite account

Results below.

## Regular gmail.com:
Unread-fix confirmed working. 
Storage_info() doesn't work though. (Output is all zeros)
Tested with English US locale.

## OG, grandfathered "Google Apps" account.
Unread-fix confirmed working.
Storage_info() doesn't work though. (Output is all zeros)
Tested with Norwegian and English UK locale.

## Paid, company GSuite account:
Unread-fix partially working. Better than before, but produces console-warnings for all counters not Inbox, and provides `0` as result, even for labels which has unreads.
Storage_info() doesn't work though. (Output is all zeros)

josteink:
I see storage_info() is implemented using a regexp on text values.

Below are the texts Ive seen for my accounts:

* Using 0.66 GB
* 6.57 GB of 17 GB used
* 2.04 GB of 15 GB used

Based on the regexp used, Id say the reason for the failure is assuming "X of Y" format and assuming percentage at the end.

Maybe you can relax the conditions slightly?

stevenirby:
@josteink thank you so much for testing! 

One big issue is that a lot of data isn't in the DOM until the page is fully loaded. After `gmail.observe.on('load', () => {})` is done. I believe the correct fix is to add more watchers to wait until more of the DOM is loaded?

> Unread-fix partially working. Better than before, but produces console-warnings for all counters not Inbox, and provides 0 as result, even for labels which has unreads.

Do you think it should return undefined instead? I didn't want to break any scripts out there using this.

> Based on the regexp used, Id say the reason for the failure is assuming "X of Y" format and assuming percentage at the end.
>Maybe you can relax the conditions slightly?

Will do.

josteink:
What might help make things easier for you is to separate data-fetching from data-processing into 2 different functions.

That way you can unit-test the processing and just add my real world examples above as test-case.

That may sound complicated at first, but I think it will help you solve this faster and better.

src/gmail.js
```diff
@@ -338,7 +338,7 @@ var Gmail = function(localJQuery) {
     };
 
     api.dom.inbox_content = function() {
-        return $("div[role=main]:first");
+        return $("div[role=main]:first") && $("[aria-label='Gmail']").length;
```
josteink, line 341: I dont understand this change. Previously the function would return a jQuery element (as formalized in the `gmail.d.ts` typescript definitions), but now you are returning a number?

The documentation also clearly states that this function should return DOM content.

```
DOM

These methods return the DOM data itself

gmail.dom**.inboxes()**
gmail.dom**.inbox_content()**
```

So what's the idea behind this change?

src/gmail.js
```diff
@@ -409,7 +409,7 @@ var Gmail = function(localJQuery) {
 
     api.get.storage_info = function() {
         var dom = $("[role=contentinfo] a").first();
-        var matches = dom.text().trim().match(/(\d+[.,]\d+) GB of (\d+) GB\s+\((\d+)%\)/);
+        var matches = dom.text().trim().match(/(\d+[.,]\d+) GB of (\d+) GB\s+\((\d+)\)/);
```
josteink, line 412: This regex will still only work for english locales, and wont be applicable to non-english users.

Please see earlier feedback on this point.

Issue #779, opened 2024-02-28
Gmail change for innerHTML


DiegoMMR:
Gmail will change and add a trustedHTML policy that doenst allow jquery to load

https://workspaceupdates.googleblog.com/2024/01/extending-trusted-types-to-gmail.html

![Untitled](https://github.com/KartikTalwar/gmail.js/assets/10916340/ae2b92e5-193e-4901-801d-8757f4009ab0)

to test it if you dont have that change you need to install this extension https://chromewebstore.google.com/detail/modheader-modify-http-hea/idgpnmonknjnojddfkpgkljpfnnfcklj

then you need to add this policy 
```
require-trusted-types-for = 'script'
```

![image](https://github.com/KartikTalwar/gmail.js/assets/10916340/c2c9d87d-9c4d-4ed4-8c48-ec9fbf06d301)

we did a change from `innerHTML` to `setHTML` is a temporal fix that worked but prob is not permament

https://developer.mozilla.org/en-US/docs/Web/API/Element/setHTML

but in theory the change to jquery 4 should fix it 

https://blog.jquery.com/2024/02/06/jquery-4-0-0-beta/

josteink:
Eeek. That sounds like it can cause quite a bit of problems, depending on the complexity of the extension.

Would you be willing to upstream and PRs which helps alleviate the issue for now?

While normally I've tried to avoid breaking changes in Gmail.js, given how this sounds like a fairly breaking change on Google's end, I might be willing to accept PRs which includes breaking changes, if that is required to solve this problem.

josteink:
Talking about how upgraded jQuery solves this... Ive considered for quite a while how it may be time to embrace regular DOM APIs instead of relying on jQuery.

Would using plain DOM APIs help us in this case? Or would we just get the same problems none the less?

DiegoMMR:
> Eeek. That sounds like it can cause quite a bit of problems, depending on the complexity of the extension.
> 
> Would you be willing to upstream and PRs which helps alleviate the issue for now?
> 
> While normally I've tried to avoid breaking changes in Gmail.js, given how this sounds like a fairly breaking change on Google's end, I might be willing to accept PRs which includes breaking changes, if that is required to solve this problem.

we have the local files for jquery and gmail.js in our repo... so what we did for now was replace all the `innerHTML` for `setHTML` with vscode and for that that worked but bc our extension was a bit big it had that innerHTML in more places but the first place where the error was show was on jquery

[...]

PR #780, opened 2024-02-29
fix: avoid using deprecated jQuery functions to prepare for jQuery 4 …

onestep:
…(#779)

- some DOM check functions were rewritten to use native DOM API
- jquery 3 dependency was removed from package.json


huksley:
Afaik based on https://blog.jquery.com/2024/02/06/jquery-4-0-0-beta/ invocation of `$.each` is not deprecated. 
I tried to analyze `gmail.js` and as far as I can see, there is only $.isArray() invocation need to be replaced. 

Would it make sense to simplify this PR so it only replaces $.isArray with Array.isArray?
With such change gmail.js fine with jQuery 4 beta in my local env.

src/gmail.js
```diff
@@ -517,37 +513,37 @@ var Gmail = function(localJQuery) {
...
 
     api.check.is_priority_inbox = function() {
-        return $(".qh").length > 0;
+        return document.querySelector(".qh") !== null;
     };
 
 
     api.check.is_rapportive_installed = function() {
-        return $("#rapportive-sidebar").length === 1;
+        return document.querySelector("#rapportive-sidebar") !== null;
```
josteink, line 521: Some of these queries check for number of results the query returns.

Would it make more sense to use `document.querySelectorAll()` instead to replace those calls, to avoid having to change the Gmail.js-side "business"-logic in these helper methods?

src/gmail.js
```diff
@@ -2378,7 +2374,7 @@ var Gmail = function(localJQuery) {
 
         // loop through applicable types
         var types = type ? [ type ] : [ "before", "on", "after", "dom" ];
-        $.each(types, function(idx, type) {
+        types.forEach(type => {
```
josteink, line 2377: If we are replacing `$.each` with something else, I would rather we replace it with a normal Javascript `for..of`-iterator if possible.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

src/gmail.js
```diff
@@ -2407,15 +2403,15 @@ var Gmail = function(localJQuery) {
     api.observe.trigger = function(type, events, xhr) {
         if(!events) return false;
         var fired = false;
-        $.each(events, function(action,response) {
+        Object.entries(events).forEach(([action, response]) => {
```
josteink, line 2406: Not saying this is wrong per se, but how did you determine which call could have `$.each` be replaced by  simple `array.forEach()` vs in some cases `Objects.keys().forEach()` and some other cases `Objects.entries().forEach()`. It seems to me like something which could be a possible source of accidental errors during the conversion.

And a follow up question: Would we avoid those issues if we instead tried to mimick the semantics of the original call by using `for..of` ?

josteink:
VERY nice work in this PR. This is taking Gmail.js in a general direction I'm very much agreement about, and I'm glad to see someone having the initiative to make it happen. Thank you VERY much!

I've left some comments in the PR code, which I think is worth considering or addressing, but have no doubt, that this is a PR I would like to see landed.

Let's do this!

src/gmail.js
```diff
@@ -18,13 +18,8 @@ var Gmail = function(localJQuery) {
...
     } else {
-        // try load jQuery through node.
-        try {
-            $ = require("jquery");
-        }
-        catch(err) {
-            // else leave $ undefined, which may be fine for some purposes.
-        }
+        // leave $ undefined, which may be fine for some purposes.
+        console.warn("GmailJS: jQuery is not provided in constructor parameters, DOM functions may not work.");
```
josteink, line 22: I see what you're trying to do here, but I think we want to be even more explicit.

I think this early in the transition, opting out from jQuery should be something that requires extra effort, and ending up without jQuery is not something which can be done accidentally.

how about....

````javascript
var Gmail = function(localJQuery) {
  var $;
    if (typeof localJQuery !== "undefined") {
        $ = localJQuery;
    } else if (typeof jQuery !== "undefined") {
        $ = jQuery;
    } else if (typeof localJQuery === "boolean" && localJQuery === false} {
        // leave $ undefined, which may be fine for some purposes.
    } else {
        throw Error("GmailJS requires jQuery to be provided as a constructor argument!");
    }
}
````

**Comment by huksley** [comment]
Btw, regarding the trusted types, as I can understand from the jQuery changes, users of jQuery need to provide "trusted" HTML, i.e. ´$(el).html(trustedHtmlHereInChromeOnly)´, jQuery will not do it automatically.

So it needs to be done on the gmail.js side.

**Comment by josteink** [comment]
> Btw, regarding the trusted types, as I can understand from the jQuery changes, users of jQuery need to provide "trusted" HTML, i.e. `$(el).html(trustedHtmlHereInChromeOnly)`, jQuery will not do it automatically.
> 
> So it needs to be done on the gmail.js side.

And it needs to be dispatched like that only when in Chrome... Sounds like a new internal helper-method?

And then if we implement that anyway... We might as well not use jQuery for the task at all?

File: src/gmail.js (hunk 1 above)
Comment (onestep) [reply src/gmail.js:521]: This is the only one query I can confirm which will change its behavior from the logic perspective. 🙂 I can change it to check for only one element like it was before, but I believe that it's either "rapportive installed" and you have the element, or it's not installed and no element is present at all.

File: src/gmail.js (hunk 2 above)
Comment (onestep) [reply src/gmail.js:2377]: Yep, will change.

File: src/gmail.js (hunk 3 above)
Comment (onestep) [reply src/gmail.js:2406]: I've reviewed all changed places to understand the types, and additionally (you've probably noticed 🙂) I've added JSDoc annotation to root object to associate it with d.ts, so IDE may now analyze parameters and properties:
image

Regarding the follow-up question - for-of works only for iterable objects, and `Object` is not iterable, so we would still need `Object.keys()` or `Object.entries()`, like:
```suggestion
        for (let [action, response] of Object.entries(events)) {
```

src/gmail.js
onestep: Makes sense. In that case, opting out from using jQuery would be really explicit, so we could even think of returning raw `HTMLElement`s instead of jQuery wrappers.

src/gmail.js
onestep: Changed.

src/gmail.js
onestep: Updated. Additionally, updated signatures in type definitions to match expected input.

src/gmail.js
onestep: Updated.

onestep:
Regarding TrustedHTML support, I believe this could be done like:
```
const rawHTMLPolicy = window.trustedTypes?.createPolicy('forceInner', {
	createHTML: input => input
});
...
$(el).html(rawHTMLPolicy ? rawHTMLPolicy.createHTML(html) : html)
```

src/gmail.js
josteink: Fair enough. Lets keep it as is.

src/gmail.js
```diff
@@ -17,14 +17,10 @@ var Gmail = function(localJQuery) {
         $ = localJQuery;
     } else if (typeof jQuery !== "undefined") {
         $ = jQuery;
+    } else if (localJQuery === false) {
+        // leave $ undefined, which may be fine for some purposes.
```
josteink, line 21: This is my fault really, but this code needs to be tweaked slightly.

We need this check before `if (typeof localJQuery !== "undefined") {` otherwise this code will never trigged.

josteink:
Simply amazing. Very nice!

One slight nitpick left (which is my fault, really)... But apart from that, this is looking great.

That said, this version may not be a drop-in replacement for existing versions without risk of breaking... So based on that I guess this should be a 2.0 release (following SemVer conventions).

Do you want me to push a non-breaking fix for the reply-menu thing we just merged first, or should we take them both in one go? I'm open to both options.

Also: We definitely need a changelog-item for this. You fix? :smile:

josteink:
Also: Thinking about loud the elimination of jQuery as a dependency completely (and thus from our API), I think it could be achieved in a fairly non-breaking way, by accepting an "output-transformer" type parameter.

Said in TypeScript, roughly like this:

````typescript
type Transformer<TElem extends HTMLElement, TResult> = (el: TElem) => TResult;
declare class Gmail<Transformer> { ... };

const gmailNoJquery = new Gmail(); // implied identity transformer (i => i)
const element = gmailNoJquery.dom.compose().$el; // HTMLElement

const gmailJquery = new Gmail(el => $(el));
const element = gmailNoJquery.dom.compose().$el; // JQuery
````

That exact code might not compile, but you get the general idea.

Basically jQuery is something you need to bring if you want it, and if you do, you get to keep 100% the API you used to have.

**Comment by onestep** [comment]
Hello @josteink,

I believe that this version does not introduce any breaking API changes, for now it would operate as before - jQuery wrappers would be in place, and if consumer has `jQuery` object in global scope, it would be used. The only change is that jQuery-less mode is made explicit, but I think it's not a frequent use case.

Let's keep 2.0 for the version with element transformers, or even jQuery-less. 🙂 For now I'd say that this PR is mostly a hotfix for jQuery 4 compatibility.

I've updated constructor argument check order and added a CHANGELOG entry.

**Comment by onestep** [comment]
Regarding your idea with transformers - let's continue in issue #592 discussion. In general, I think that transformers for input and output and for single/multiple elements would be needed - i.e. `dom.inbox_content()` returns single element, while `dom.inboxes()` returns multiple. Also, `check.is_peoplekit_compose(el)` is assumed to expect `el` as a single element, while you can pass multiple elements wrapped to `$`. 🙂 At some point I could think that gmail-js 2 should be a complete jQuery-less spinoff, while gmail-js 1 with jQuery support could stay in security/hotfix maintenance mode for a transition period.

File: src/gmail.d.ts [hunk 6]
```diff
@@ -1031,14 +1034,14 @@ interface GmailCache {
 ////////////////////////////////////////////////////////////////////////////////
 
 declare class Gmail<T extends string=never> {
-    constructor(localJQuery?: JQueryStatic);
+    constructor(localJQuery?: JQueryStatic | false);
```
huksley, line 1037: In my opinion, it would best to make this parameter mandatory ´jQuery | false´

That way, typescript code which were calling new Gmail()
will fail to compile with the new version, instead of throwing the runtime exception.

josteink: Fixed

onestep: In that case we would probably need to remove check for global jQuery in code (as we either would provide it or want to disable it) and *that* would break compatibility. 🙂
Or it could be OK to have TypeScript definitions stricter than actual library expectation - not everyone uses TS in the end.

josteink:
I've made some very small ammendments based on comments from @huksley. I've also tested some of the changes which I can test, and they seem good.

The ones I couldn't test was because of `TrustedHTML` now being enabled on all accounts I've tried, and that breaking the changed code.

Based on that, I think it is safe to merge this, publishing, and then we can start working on solving the bigger issues.

Thanks for all the work so far, guys!

